PR #81, opened 2015-08-30
Support dsd disc images via UI and URL params

ribbons:
Internally, the `baseSsd` method in fdc.js supports dsd images if the `dsd` member is true, however this isn't exposed in the UI or via URL parameters.

This PR adds transparent handling of a disc image as dsd if it has the extension `.dsd` and is loaded via a URL parameter or using the UI to load a local image.

What I'm not sure about is if the functions I've changed (`ssdFor` and `baseSsd`) should be renamed to reflect that they don't just support SSDs - e.g. `diskFor` and `baseDisk`, or if they should be left as-is for compatibility?  Let me know if they should be renamed and I'll update this PR.

Also included is a fix for a bug running the tests via a browser if the application isn't located in the root of a web server (which I discovered along the way).


mattgodbolt:
Fantastic; thanks for this!

PR #83, opened 2015-08-31
Remove the 'content' class from outer monitor div

ribbons:
This prevents the emulated screen unexpectedly jumping to the right (and so partially off-screen) if the window size is less than 992px wide, but doesn't otherwise affect the layout.


mattgodbolt:
Thanks :)

PR #89, opened 2015-10-03
Video changes and 'different' teletext

richtw1:
The graphics chars are programatically generated now, and the emulation is more precise than B-Em.  Takes some getting used to though, visually.  They seem a bit bold, but that's what they were actually like I think!  See what you reckon.

Now no hacks for Mode 7, display skew is emulated correctly.  Plus various fiddles and bugfixes.  And interlace not annoying any more.


murkle:
Travis build has failed with:

        throw err;
              ^
Error: Evaluating ./video.js as module "video" failed with error: SyntaxError: Use of const in strict mode.

mattgodbolt:
I fixed the tests in a branch and all's well there. Looks awesome @richtw1 :) I'm gonna merge this then apply all the fixes shortly. Some comments inline

murkle:
Revs seems to have some extra pixels where the blocks meet - no idea if that's what you get on a real BBC though :)

http://
revs

http://bbc.godbolt.org/beta/?disc=sth%3AAcornsoft%2FRevs.zip&autoboot
http://bbc.godbolt.org/?disc=sth%3AAcornsoft%2FRevs.zip&autoboot

richtw1:
Good catch.  I need to disable character rounding on graphics characters (SAA5050 datasheet says it only does it for alphanumeric characters).

Another other thoughts on the new Mode 7 glyphs?  I still find them a bit jarring on first sight but get used to them quickly - I think I got too used to the B-Em way of rendering them.  But they are rendered exactly as they are generated by the hardware.  I think maybe we should just add a TV effect shader to recreate the correct visuals, rather than trying to modify the teletext glyphs.

murkle:
I think they are much better than the old ones, nice and clear

The green dots at the bottom of Eng_test.ssd and the dots in row 02 are now evenly spaced as required.
http://stardot.org.uk/forums/viewtopic.php?t=4686

real_bbc

Also just noticed a bug in the old version - the digits in the last 2 columns seem pixelated. Now fixed :)

richtw1:
For comparison, here's the test running on jsbeeb now:
image
Comparing with a real Beeb screenshot, I think the glyphs are spot on now too.  The only issues I can see are the absence of the cyan and white stripes at the top right, and I'm not sure whether we should be able to see 'Flash' or not?  (or whether 'Box' should be flashing?)

richtw1:
And for extra comparison, here's what it _used_ to look like:
image
New code is definitely an improvement I think, based on this screenshot at least.

murkle:
Pressing <Space> makes the Cyan and White stripes appear (and Conceal/Flash at the bottom)

richtw1:
Ah great!  So this is 100% correct then I think :)

murkle:
(consistent with http://stardot.org.uk/forums/download/file.php?id=3008 so all correct I think)

PR #110, opened 2016-03-01
Minor LED panel fixes

ribbons:
Just a couple of small commits to fix two minor LED panel niggles:

The first makes the background of the LED panel fully opaque - currently there is a transparent inner border between the `div` border and `table` which doesn't show up on the dark background, but is visible sometimes with a custom footer.  I've set a partially white background in the below before and after for maximum visibility:

before after

The second evens up the spacing around the LED images when the text is hidden on small screens or in an `iframe` - here's another before and after:

before_vert after_vert

### DIFF
--- a/jsbeeb.css
+++ b/jsbeeb.css
@@ -110,6 +110,10 @@ span.accesskey {
     #leds thead {
         display: none;
     }
+
+    div.led {
+        vertical-align: middle;
+    }
 }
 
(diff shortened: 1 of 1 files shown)


mattgodbolt:
Thanks for this!

mattgodbolt:
also pushed to live site. thanks again!

PR #145, opened 2017-05-19
Fix some teletext characters.

tom-seddon:
I noticed some of the teletext characters weren't quite right, after comparing the font with the display from my Master 128.

* 1/4 and 1/2 fill the bottom row (the 1 is 5 pixels high, like the 3 in 3/4)
*  _ and - are aligned with the horizontal lines in + and divide
* M has an extra pixel in the middle

Here's a closeup of these symbols on my Master 128:

image

(The blocks below the 1/4, 1/2 and 3/4 symbols are graphics mode CHR$255 blocks)

Thanks,

--Tom


mattgodbolt:
You star! Thanks so much @tom-seddon !

mattgodbolt:
My autoupdate is broken atm so it'll take a while before I push out the fix to the live site but yay!

PR #147, opened 2017-06-18
Allow loading of remote tape images

pau1ie:
An undocumented feature is that a disc image can be loaded from a URL. This isn't the case for tape images. This modification allows a tape to be downloaded from a URL. 

See the [Stardot Forums](http://stardot.org.uk/forums/viewtopic.php?f=4&t=13272) for more discussion. 

I only did a quick sanity check that the modification works to download a remote image, and also disc images still work.


mattgodbolt:
Thanks for this!

PR #157, opened 2017-11-26
Fix hold mosaics processing.

tom-seddon:
When hold mosaics mode is off, control codes reset the held char.

As noted in http://www.stardot.org.uk/forums/viewtopic.php?f=4&t=14105


mattgodbolt:
You star! Thanks Tom.

I need to add a test to compare a "known good" teletext image with the engineering sample... (filed as #159 now)

PR #170, opened 2018-07-09
Swapped key codes for mac users.

aidansean:
For Mac users, the key codes for APOSTROPHE and BACK_QUOTE seem to be the wrong way around compared what jsbeeb expects.  This makes it almost impossible to play games where the BBC.COLON_STAR key is used for “up”.  This fix checks to see if the user has a Mac operating system and if so, swaps the key codes around so that they match what jsbeeb expects.

### DIFF
--- a/utils.js
+++ b/utils.js
@@ -290,6 +290,14 @@ define(['jsunzip', 'promise'], function (jsunzip) {
         keyCodes.BACK_QUOTE = isUKlayout ? 223 : 192;
     }
 
+    // Swap APOSTROPHE and BACK_QUOTE keys around for Mac users.  They are the opposite to what jsbeeb expects.
+    // Swap them to what jsbeeb expects, and tidy up the hash key to prevent duplicate key mappings.
+    if (window && window.navigator.userAgent.indexOf("Mac") !== -1) {
+        keyCodes.BACK_QUOTE = 192;
+        keyCodes.APOSTROPHE = 222;
+        keyCodes.HASH = 223;
+    }
+
     exports.getKeyMap = function (keyLayout) {
         var keys2 = [];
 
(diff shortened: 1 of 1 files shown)


mattgodbolt:
Hi! Unfortunately the tests are now failing, e.g. https://travis-ci.org/mattgodbolt/jsbeeb/jobs/401818313

A relatively simple check to see if "window" exists ought to fix them: a lot of the code runs in a non-browser environment for tests.

utils.js
```diff
@@ -556,6 +556,12 @@ define(['jsunzip', 'promise'], function (jsunzip) {
             map(keyCodes.BACK_QUOTE, BBC.AT);
         }
 
+        // Swap APOSTROPHE and BACK_QUOTE key codes around for Mac users.
+        if (window.navigator.userAgent.indexOf("Mac") !== -1) {
```
mattgodbolt: As noted elsewhere, maybe change to

```
if (window && window.navigator.....) {
```

(NB the space between the `if` and `(` too, just to be in keeping with the rest of the codebase.

aidansean: Cool, I'll get this done shortly.  Sorry for the slow reply, I've had a busy few days :)

aidansean:
Changes made!  I spent some more time thinking this change over and realised that the best place to change the mappings was upstream of the various choices of layout.

Checks still failed.  I'll take a look into that. I might just need to add another block to test for the existence of `window` in its own clause.

aidansean:
node is having problems with the `window` variable.  I'm afraid I don't know much about node, so I'll leave this with you for now, and do some reading up on the subject over the weekend.

utils.js
```diff
@@ -292,10 +292,12 @@ define(['jsunzip', 'promise'], function (jsunzip) {
 
     // Swap APOSTROPHE and BACK_QUOTE keys around for Mac users.  They are the opposite to what jsbeeb expects.
     // Swap them to what jsbeeb expects, and tidy up the hash key to prevent duplicate key mappings.
-    if (window && window.navigator.userAgent.indexOf("Mac") !== -1) {
-        keyCodes.BACK_QUOTE = 192;
-        keyCodes.APOSTROPHE = 222;
-        keyCodes.HASH = 223;
+    if (window) {
```
mattgodbolt: Ah good point. I'll fix it up with some Google-fu :grinning:

mattgodbolt:
Thanks: I'll merge it and then fix the error :) Sorry...JS is simply the worst language...

aidansean:
Yay!  Thanks, now I can play Citadel and Labyrinth again :D

mattgodbolt:
I'll update the live site (once I remember how...)

mattgodbolt:
Done!

PR #177, opened 2018-10-06
First working asm.js version

mattgodbolt:


ihh:
nice

mattgodbolt:
It was fun but somewhat against the spirit of jsbeeb I think, so...closing.

PR #191, opened 2019-07-27
Fix VIA PB7 handling, hit by Snapper.

### LINKED ISSUE #185: Snapper freezes at startup
https://bbc.godbolt.org/?disc=sth%3AAcornsoft%2FSnapper-v2.zip&autoboot#

```
10a5ad 60 fe.`. LDA $fe60
10a8cd 60 fe.`. CMP $fe60
10abf0 fb.. BEQ $10a8
```

Hangs here, with `A=ff`

(from #179)

scarybeasts:
Fixes #185 

VIA PB7 handling behavior was pinned down reasonably thoroughly in these Stardot threads:
https://stardot.org.uk/forums/viewtopic.php?f=4&t=16081
https://stardot.org.uk/forums/viewtopic.php?f=4&t=16263

I'll attach the tests.ssd file to the issue. Both PB7 tests (VIA.PB7 and VIA.PB2) now pass with the same results as observed on a real beeb. Perhaps more cause for celebration is that Snapper now works fine. Planetoid also pulls tricks in this department and still works with the changes applied.

### DIFF
--- a/via.js
+++ b/via.js
@@ -35,6 +35,7 @@ define(['./utils'], function (utils) {
             porta: 0, portb: 0,
             ca1: 0, ca2: 0,
             justhit: 0,
+            t1_pb7: 0,
 
             reset: function (hard) {
                 // http://archive.6502.org/datasheets/mos_6522_preliminary_nov_1977.pdf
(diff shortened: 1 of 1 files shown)


mattgodbolt:
Thanks again!

PR #194, opened 2019-07-28
Initialize RAM to 0x00, not 0xFF.

### LINKED ISSUE #105: Clogger hangs if run from cold boot
From a beeb image: http://bbc.godbolt.org/?disc=sth%3AImpact%2FClogger.zip# hangs (hitting a HLT at 0x2a2e seemingly executing some ASCII...)

Works fine on Master, and works from a warm boot.

scarybeasts:
Fixes #105 

Clogger relies on &9E being zero initialized, otherwise it fails.

Let's just zero initialize RAM on hard rest instead of setting it to 0xFF. This likely fixes other things too.

### DIFF
--- a/6502.js
+++ b/6502.js
@@ -858,7 +858,7 @@ define(['./utils', './6502.opcodes', './via', './acia', './serial', './tube', '.
                         }
                     }
                     for (i = 0; i < this.romOffset; ++i)
-                        this.ramRomOs[i] = 0xff;
+                        this.ramRomOs[i] = 0x00;
                     this.videoDisplayPage = 0;
                     this.scheduler = new scheduler.Scheduler();
                     this.soundChip.setScheduler(this.scheduler);


chrisn:
Initialising RAM to 0xFF was added to fix a bug in a different game. On power up, my Beeb reports:
```
>P. ?&9E
       255
```
See [this post](https://stardot.org.uk/forums/viewtopic.php?f=7&t=4927&start=120#p102365) and replies for more details. Essentially, the RAM powers up in an uninitialised state, and the OS doesn't zero out all the addresses on startup, so the values may vary between machines.

scarybeasts:
Thanks @chrisn. Very interesting.

Let me see if I can find some stardot community members to see if this Clogger image works on a real beeb at cold start.
If we want both games to work on cold boot, we could do something cheesy like zero initialize the zero page only...

scarybeasts:
We can close this and consider https://github.com/mattgodbolt/jsbeeb/pull/195.

scarybeasts:
Closing.

PR #195, opened 2019-07-29
Initialize zero page to 0x00.

### LINKED ISSUE #105: Clogger hangs if run from cold boot

scarybeasts:
Fixes #105 

Clogger relies on &9E being zero initialized, otherwise it fails.
Conversely, Eagle Empire is known to rely on &800 being initialized to 0xFF.

We are free to initialize DRAM however we please because it is not defined.

This patch keeps both Clogger and Eagle Empire happy.

### DIFF
--- a/6502.js
+++ b/6502.js
@@ -857,8 +857,18 @@ define(['./utils', './6502.opcodes', './via', './acia', './serial', './tube', '.
                             this.memLook[i] = this.memLook[256 + i] = 0;
                         }
                     }
-                    for (i = 0; i < this.romOffset; ++i)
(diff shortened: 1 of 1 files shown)


mattgodbolt:
Seems rather suspicious to have to special case RAM, but I don't have a better idea, and as you say we can make up whichever values we like!

PR #196, opened 2019-07-29
Speech chip is not present so don't interfere with data bus

### LINKED ISSUE #93: Keyboard not responding in some games.
Repton 2 (alt version) in STH archive hangs on title screen.
http://bbc.godbolt.org/?disc=sth%3ASuperior%2FRepton2-alt.zip&autoboot

Works fine on B-Em and BeebEm.

Interesting point: even the CAPS LOCK LED doesn't work. Weird.

scarybeasts:
Fixes #93 

For some reason, the alternative Repton 2 images are latching the speech chip read enable (active low):
[ITRP] 4160: SEI
[ITRP] 4161: LDA #$01
[ITRP] 4163: STA $FE40

I don't think this should do anything since we're not emulating the speech chip. Surely a chip that isn't present shouldn't be able to affect the slow data bus?

The removed code includes a comment "no speech" which raises the question: is the speech chip still correctly detected as not present? The answer is yes. This is handled by self.readPortB returning the high bit set (it returns 0xFF). This can be confirmed with:
P. ?&27B
Which returns 0 for no speech chip detected by MOS.

Simplified behavior matches b-em.

### DIFF
--- a/via.js
+++ b/via.js
@@ -486,12 +486,6 @@ define(['./utils'], function (utils) {
             if (!(self.IC32 & 8) && !self.keys[self.keycol][keyrow]) {
                 self.sdbval &= 0x7f;
             }
-            if (!isMaster && !(self.IC32 & 4)) {
-                self.sdbval = 0xff; // unsure; taken from beebem
-            }
-            if (!isMaster && !(self.IC32 & 2)) {
-                self.sdbval = 0x00;  // no speech
-            }
         };
 
         self.writeIC32 = function (val) { // addressable latch


via.js
```diff
@@ -486,12 +486,6 @@ define(['./utils'], function (utils) {
             if (!(self.IC32 & 8) && !self.keys[self.keycol][keyrow]) {
                 self.sdbval &= 0x7f;
             }
-            if (!isMaster && !(self.IC32 & 4)) {
```
mattgodbolt, line 489: Anything that removes suspicious code I copied from other emulators without understanding...

PR #197, opened 2019-07-30
Model sound chip write enable closer to real hardware.

### LINKED ISSUE #96: Sound error in Repton 2 Speech demo
There's some background noise during the speaking of the word "Superior" at the start of the Speech demo when loading Repton 2. B-em has the same problem but it sounds ok in BeebEm.

### LINKED ISSUE #193: Speech.dsd speech demo by D.A.W. and C.J.H. doesn't play
Breaking out from #167 because there's a separate issue.

http://bbcmicro.co.uk/gameimg/discs/1411/Speech.dsd
doesn't play the sound correctly in jsbeeb. It's because it repeatedly toggles on sound write enable without ever toggling it off, and the emulators want to see an on -> off transition before they send a byte of data over to the sound chip.

Looks like a real beeb is more permissive than that.

scarybeasts:
Fixes #96 
Fixes #193 

There's a great, concise summary of our latest understanding of hardware behavior from Rich here:
https://stardot.org.uk/forums/viewtopic.php?f=4&t=17537#p243442

This patch takes us closer to that behavior while being minimally invasive. I think it's a good balance for now and I no longer know of any sampled (or other) sound that jsbeeb cannot play.

Now works:
Speech.dsd
Repton 2

Tested still ok:
Exile
Enjoy the Silence
Spy Hunter
Reet Petite
Citadel

### DIFF
--- a/soundchip.js
+++ b/soundchip.js
@@ -239,10 +239,10 @@ define(['./utils'], function (utils) {
         this.slowDataBus = 0;
         this.updateSlowDataBus = function (slowDataBus, active) {
             this.slowDataBus = slowDataBus;
-            if (active && !this.active) {
+            this.active = active;
+            if (active) {
                 activeTask.reschedule(minCyclesWELow);
             }
-            this.active = active;
         };
         this.reset = function (hard) {
             if (!hard) return;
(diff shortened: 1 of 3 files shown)


mattgodbolt:
Wonderful! Thanks so much!

PR #200, opened 2019-08-01
Use a more compatible set of default SWRAM banks.

### LINKED ISSUE #199: Dunjunz vs. sideways RAM configuration
There are various reports of Dunjunz not working. From quickly testing all the versions referenced in various threads, I think there are two issues:

- Disc060.dsd
Expects sideways RAM in banks 4 and 5, but jsbeeb's default beeb+swram model has chosen different banks. Error message is pretty clear.

- Dunjunz-swram.ssd
Expects sideways RAM in banks 1 and 2, but jsbeeb's default beeb+swram model has chosen different banks. Uninformative error message when it fails.

Other versions of Dunjunz, including the Master version, appear to load fine.

@mattgodbolt any particular reason you chose the sideways RAM banks as you did? Should we change them to accommodate the Dunjunz images? (Before anyone says "Exile", note that it will accept sideways RAM in any position.)

scarybeasts:
Fixes #199 

Everything significant I know of works "out of the box" now. It's a lot of SWRAM but there are still 6 slots left for extra ROMs which should suffice.

Now works:
- Crazee Rider.
- Dunjunz (2x different variants with different needs).

Still works:
- Exile.
- Enjoy the Silence demo.

### DIFF
--- a/models.js
+++ b/models.js
@@ -15,10 +15,10 @@ define(['./fdc'], function (fdc) {
 
     // TODO: semi-bplus-style to get swram for exile hardcoded here
     var beebSwram = [
-        true, true, false, false,
+        true, true, true, true,      // Dunjunz variants. Exile (not picky).
+        true, true, true, true,      // Crazee Rider.
         false, false, false, false,
-        false, false, false, false,
-        true, true, false, false];
+        false, false, false, false];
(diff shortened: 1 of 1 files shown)


mattgodbolt:
Thanks!

PR #202, opened 2019-08-06
Fix various i8271 bugs including the excessive seeking bug.

### LINKED ISSUE #165: Disc drive seems to erroneously reset to track 0
Lots of seeks to and from track zero that don't seem to happen in b-em etc

scarybeasts:
Fixes #165 

Some of these bugs fixed are pretty overt so I suppose DFS-0.9 doesn't use all the controller functionality.

Bugs fixed:
- Freshly loaded disc + drive started at physical track -1.
- Incorrect value was being written to track registers.
- Report write protect bit correctly instead of always.
- Spinup wasn't always waited for.
- Spindown timer wasn't always set after an operation.
- Spindown should invalidate drive ready status.
- Remove Elite-A workaround (other fixes make it unnecessary).
- Track 0 sensor should report physical track 0 position, not logical.
- Prevent physical seeks out of bounds.
- Drive status queried incorrectly.

After all this, the drive isn't seeking to track 0 all the time, but Elite-A still loads. It relies on spindown occurring before the ship is launched, which is probably how this ever worked on a real beeb. Spindown timer set to 2 second to occur a bit faster to be conservative.

### DIFF
--- a/fdc.js
+++ b/fdc.js
@@ -91,7 +91,7 @@ define(['./utils'], function (utils) {
         this.sectorOffset = 0;
         this.formatSector = 0;
         this.rsector = 0;
-        this.track = -1;
+        this.track = 0;
         this.side = -1;
         this.notFoundTask = fdc.scheduler.newTask(function () {
             this.fdc.notFound();
@@ -218,7 +218,6 @@ define(['./utils'], function (utils) {
         self.status = 0;
         self.result = 0;
         self.data = 0;
-        self.driveSel = 0;
         self.curDrive = 0;
         self.drvout = 0;
         self.curTrack = [0, 0];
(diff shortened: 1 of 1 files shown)


fdc.js
```diff
@@ -401,51 +397,54 @@ define(['./utils'], function (utils) {
                 default:
                     self.result = self.status = 0x18;
                     self.NMI();
-                    self.callbackTask.cancel();
                     break;
             }
         }
 
         function spinup() {
-            // TODO: not sure where this should go, or for how long. This is a workaround for EliteA
```
mattgodbolt, line 410: Thanks for working this out!

mattgodbolt:
I'll merge and fix up the `lint` errors. Thanks so much!

PR #205, opened 2019-08-07
Handle light pen registers on low -> high CB2 transition.

### LINKED ISSUE #135: Is the lightpen interrupt modelled?
I tried and failed to run Pharaoh's Curse [here](http://bbcmicro.co.uk/jsbeeb/play.php?autoboot&disc=http://bbcmicro.co.uk/gameimg/discs/Disc027-PharaohsCurse.ssd) - the game hangs to a black screen at the point I'd expect it to draw the playfield.

I wonder if that's anything to do with the lightpen interrupt, as mentioned [here](http://stardot.org.uk/forums/viewtopic.php?p=32556#p32556):
"On the Beeb you can trigger a 'fake' light pen interrupt, which makes the CRTC latch the current video address into the light pen registers. Handy for synchronisation. Beebem doesn't emulate this though, only B-em and BeebinC do."

There are other tricky games mentioned in [this thread](http://stardot.org.uk/forums/viewtopic.php?p=86252#p86252).

scarybeasts:
Fixes #135 

Even without a light pen physically attached, it's possible to put system VIA CB2 into output mode such that the CRTC sees what appears to be a light pen pulse. So when this happens, set the light pen position registers.

Actually needed by Pharoah's Curse to start the game despite it being just a normal keyboard game.

### DIFF
--- a/video.js
+++ b/video.js
@@ -53,6 +53,7 @@ define(['./teletext', './utils'], function (Teletext, utils) {
         this.cursorPos = 0;
         this.interlacedSyncAndVideo = false;
         this.doubledScanlines = true;
+        this.prevcb2 = false;
 
         this.topBorder = 12;
         this.bottomBorder = 13;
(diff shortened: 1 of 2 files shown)


video.js
```diff
@@ -366,6 +367,17 @@ define(['./teletext', './utils'], function (Teletext, utils) {
         ////////////////////
         // Main drawing routine
         this.polltime = function (clocks) {
+            // Even with no light pen physically attached, the system VIA can
```
mattgodbolt, line 370: It's a shame we have to do this work on every call to `polltime()`. It's the most-called routine in the emulator. That said: I can't see any way around it, unless you have a more cunning idea, so this is 100% ok :)

scarybeasts:
One alternate way this could be arranged is to have a callback upon CB2 level change. I'll see if it looks cleaner.

PR #207, opened 2019-08-07
Don't use the CMOS data bus if address strobe is still high.

### LINKED ISSUE #152: uridium is stuck in demo mode
please see this thread
http://stardot.org.uk/forums/viewtopic.php?p=184566#p184566

(Still might be user error, but affects two people, and apparently the SSD is good on other emulators)

scarybeasts:
Fixes #152 

Uridium manages to enable the CMOS chip on a Master, but it doesn't interfere with the keyboard on real hardware. This replicates that behavior.

### DIFF
--- a/cmos.js
+++ b/cmos.js
@@ -27,7 +27,11 @@ define([], function () {
         }
 
         function cmosRead(IC32) {
-            // b-em comment is: To drive bus, CMOS must be enabled, D must be high, RW must be high.
+            // To drive the bus we need:
+            // - CMOS enabled.
+            // - Address Strobe low.
+            // - Data Strobe high.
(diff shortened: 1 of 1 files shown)


cmos.js
```diff
@@ -27,15 +28,15 @@ define([], function () {
 
         function cmosRead(IC32) {
             // b-em comment is: To drive bus, CMOS must be enabled, D must be high, RW must be high.
```
mattgodbolt: If possible, can this comment be updated, and if not, maybe removed? Now we're diverging from what b-em does the comment doesn't match the code beneath (as is so often the case with comments...)

scarybeasts: Good point, will update.

Issue #56, opened 2015-02-02
Add a "run at full speed" setting


mattgodbolt:
As in, no delay at all. Also maybe % of real time.

BigEd:
Would this speed up cassette loading? I was playing with Acornsoft Lisp which takes minutes to load - and I can't even background the tab, or it stalls!
http://bbc.godbolt.org/?tape=sth%3Alang%2FAcornsoft%2FLisp_B.zip&autochain

mattgodbolt:
Whoops: tape loading is supposed to continue even when in a separate tab. Speed tape loading is definitely something I have on my to-do list: lots of it comes down to not doing video processing as often as most of the hard work is really the video emulation: running at "full speed" on my pretty beefy laptop is still only ~2x faster in my experiments, so more needs to be done. For a laugh you can try the "cpuMultiplier" on the beta site, e.g. 
http://bbc.godbolt.org/beta/?tape=sth%3Alang%2FAcornsoft%2FLisp_B.zip&cpuMultiplier=2 (autochain appears broken at the moment so you'll need to type *TAPE etc)

BigEd:
Hmm, the beta version froze when not the front tab, and failed to unfreeze when displayed again.

mattgodbolt:
Oops; seems like I've broken some things there. Will investigate: thanks.

scarybeasts:
These days, "as fast as possible" is about 3x on my rubbish old laptop.

Rather than lowering emulation accuracy in fasttape mode, I might look to see if there are any generic speedups available in either the CPU or video code.

mattgodbolt:
That's great: thanks @scarybeasts !

PR #208, opened 2019-08-07
Add a "go fast" keyboard toggle and a fasttape URL parameter.

### LINKED ISSUE #8: Support faster-than-realtime tape loading

scarybeasts:
Fixes #8, #56 

Ctrl + INSERT now toggles "max speed" on or off.
The new URL parameter &fasttape will automatically engage "max speed" when the cassette motor is on.

In both cases, accuracy is unaffected. It's just a case that virtual time runs faster than real time, to the extend of the hardware's capabilities.

On my ancient laptop, "max speed" is about 3x.

### DIFF
--- a/main.js
+++ b/main.js
@@ -39,6 +39,8 @@ require(['jquery', 'underscore', 'utils', 'video', 'soundchip', 'ddnoise', 'debu
         var keyCodes = utils.keyCodes;
         var emuKeyHandlers = {};
         var cpuMultiplier = 1;
+        var fastAsPossible = false;
+        var fastTape = false;
 
         if (queryString) {
             if (queryString[queryString.length - 1] === '/')  // workaround for shonky python web server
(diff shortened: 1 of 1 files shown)


main.js
```diff
@@ -325,6 +330,9 @@ require(['jquery', 'underscore', 'utils', 'video', 'soundchip', 'ddnoise', 'debu
             } else if (code === utils.keyCodes.HOME && evt.ctrlKey) {
                 utils.noteEvent('keyboard', 'press', 'home');
                 stop(true);
+            } else if (code === utils.keyCodes.INSERT && evt.ctrlKey) {
+                utils.noteEvent('keyboard', 'press', 'insert');
```
mattgodbolt, line 334: Thanks for keeping stats here :) (not that I've looked at them in years...but will do!)

main.js
```diff
@@ -1192,13 +1200,26 @@ require(['jquery', 'underscore', 'utils', 'video', 'soundchip', 'ddnoise', 'debu
                 last = 0;
                 return;
             }
-            requestAnimationFrame(draw);
+            var speedy;
```
mattgodbolt: maybe rephrase as:

```javascript
var speedy = fastAsPossible || (fastTape && processor.acia.motorOn);
if (speedy) { ... }
```

Seeing an uninitialised variable tickles my C++ spider sense (even though it's totally fine in this context :))

scarybeasts: Fixed in latest upload.

Issue #166, opened 2018-02-21
Simulated Shift keypress on autoboot is held for too long, disrupts paged scrolling


ahope1:
If you click on an autobooting JSBeeb link [like this one](https://bbc.godbolt.org/?autoboot&disc=https://dl.dropboxusercontent.com/s/f7zpglqfawowlqd/VDU14bug.ssd), which has a !BOOT that CHAINs a BASIC program that immediately uses VDU14 to activate paged scrolling mode (which requires the user to press Shift to scroll the screen in the emulated Beeb) before printing an infinite list of numbers, then the simulated Shift keypress (which I assume is JSBeeb's way of implementing a Shift-Break to boot the disc-image) is held for too long and causes the screen to scroll a couple of times when it shouldn't:

[https://bbc.godbolt.org/?autoboot&disc=https://dl.dropboxusercontent.com/s/f7zpglqfawowlqd/VDU14bug.ssd](https://bbc.godbolt.org/?autoboot&disc=https://dl.dropboxusercontent.com/s/f7zpglqfawowlqd/VDU14bug.ssd)

Try pressing Shift-Break to boot the disc-image manually to see the correct behaviour.

More details here:

[http://www.stardot.org.uk/forums/viewtopic.php?f=4&t=14581#p194938](http://www.stardot.org.uk/forums/viewtopic.php?f=4&t=14581#p194938)

murkle:
Is this a problem given that " I've chosen to rework my program (an instructions displayer) and eliminate scrolling altogether!"

Suggested solution of checking for disc access doesn't seem very robust (eg with no DFS I think Shift-break works for Tape Filing System and (??) Rom Filing System)

ahope1:
@murkle It's a problem in that it's not the behaviour that a user would expect. Also, not every disc ever made since the 1980s can easily be changed to work around this bug!

You can't Shift-Break a tape. Don't know about ROM FS. 

There's another suggested solution further down the thread:

http://www.stardot.org.uk/forums/viewtopic.php?f=4&t=14581&sid=cf775515a089016e56c624ee05b2dbb3#p194947

richtw1:
In OS 1.20, the check for Shift held on reset happens at &DB87. The call to
set the keyboard LEDs calls KEYV to do a direct check for Shift and Ctrl
held (which is maintained in &25A).  Anyway, point is, Shift only needs to
be held high on autoboot until then.  I imagine 1/10th of a second after it
[...]

PR #211, opened 2019-08-12
Add a noseek parameter which makes disc seeking instant.

scarybeasts:
The bbcmicro.co.uk admins asked for this. They currently have a noiseless and instant disc seek because they have an old jsbeeb. This UX is deemed preferable by them so they can't readily upgrade their embedded jsbeeb copy.

Since it's pretty trivial to add a noseek parameter to get the old behavior, this patch does so.

### DIFF
--- a/ddnoise.js
+++ b/ddnoise.js
@@ -125,7 +125,7 @@ define(['./utils', 'underscore', 'promise'], function (utils, _) {
     FakeDdNoise.prototype.seek = function () {
         return 0;
     };
-    FakeDdNoise.prototype.initialize = function () {
+    FakeDdNoise.prototype.initialise = function () {
         return Promise.resolve();
     };
 
--- a/fdc.js
+++ b/fdc.js
@@ -315,7 +315,7 @@ define(['./utils'], function (utils) {
             return self.data;
         };
         self.discFinishRead = function () {
-            self.callbackTask.reschedule(200);
+            self.callbackTask.reschedule(DiscTimeSlice);
         };
 
         var paramMap = {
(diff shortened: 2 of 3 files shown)


ddnoise.js
```diff
@@ -125,7 +125,7 @@ define(['./utils', 'underscore', 'promise'], function (utils, _) {
     FakeDdNoise.prototype.seek = function () {
         return 0;
     };
-    FakeDdNoise.prototype.initialize = function () {
+    FakeDdNoise.prototype.initialise = function () {
```
mattgodbolt, line 128: Yay! I've been in the US too long...

lurkio:
Someone should probably check that the new "noseek" mode doesn't unintentionally trigger that old scrolling bug again (since the cause of that scrolling bug seems to have been something to do with seek times (which I don't fully understand, admittedly)): https://github.com/mattgodbolt/jsbeeb/issues/166

lurkio:
Btw, will this be going live at bbc.godbolt.org? It would be handy for testing various disc-images in the immediate term.

scarybeasts:
@mattgodbolt Actually, this would be a good time to "checkpoint" and push the latest git master to https://bbc.godbolt.org/. All my recent fixes and improvements are merged, and I'm away for a few days.

mattgodbolt:
Done! I plan on trying to make this more automated too so it gets done automatically (maye on a "release" branch", or maybe just on every commit anyway...)

lurkio:
Thanks, @mattgodbolt!

@scarybeasts At first glance, the new noseek option seems to work really well. Thanks!

PR #212, opened 2019-08-13
Fix interrupt check timing on taken branches for 6502.

scarybeasts:
This patch opts for maximum accuracy including the dual poll points of a branch when it is taken with a branch crossing -- even if it's going to be difficult to observe the first with standard BBC hardware.

Kevin Edwards protection continue to run fine.

### DIFF
--- a/6502.js
+++ b/6502.js
@@ -107,9 +107,26 @@ define(['./utils', './6502.opcodes', './via', './acia', './serial', './tube', '.
                 var newPc = (cpu.pc + offset) & 0xffff;
                 var pageCrossed = !!((cpu.pc & 0xff00) ^ (newPc & 0xff00));
                 cpu.pc = newPc;
-                cpu.polltime(pageCrossed ? 3 : 2);
-                cpu.checkInt();
-                cpu.polltime(1);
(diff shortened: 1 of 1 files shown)


richtw1:
Seems fine to me.  I originally imagined implementing two functions - branchCMOS() and branchNMOS() - and generating the appropriate one in makeCpuFunctions(), but I doubt it makes any difference, with branch predictors these days, etc. (Also it's a bit of a pain, as getOp() doesn't know if65c12 or not).

When I put together a cycle-level 6502 state machine, I also had NMOS branches checking for IRQs twice (in T0 as always; and an extra one in T2, because T0 is skipped if branch taken crossing page boundary), and can confirm that Visual6502 - and hence a real 6502 - operates like this.

mattgodbolt:
Re: making two functions, JS JITs will "notice" that `model.nmos` is a constant and almost certainly fold out the "other" side of the branch, so I'm not too worried about that. And if they don't, then yeah, BPUs will make it cheap(ish) (the actual lookup and comparison will dominate I'd guess in that case). But, wow. Thanks again so much @scarybeasts

PR #214, opened 2019-08-17
Enable browser sound upon key or mouse down, plus warning message.

### LINKED ISSUE #184: Chrome needs interaction to start audio
Latest chromes complain about not being able to start audio until an interaction.

```
The AudioContext was not allowed to start. It must be resumed (or created) after a user gesture on the page. https://goo.gl/7K7WLu
```

### LINKED ISSUE #186: No sound in Safari browser on Mojave
As per the subject. JSBeeb produces no sound of any kind in Safari v12.1.1 on macOS Mojave v10.14.5.

Sound works on Mojave in Firefox though.

scarybeasts:
Fixes #184 
Fixes #186 

Hook user interactions, and enable audio.
If audio is detected as suspended, fade in a warning UI and fade it out again when the audio state changes to running.

Tested in Chrome and Firefox.

### DIFF
--- a/ddnoise.js
+++ b/ddnoise.js
@@ -55,7 +55,7 @@ define(['./utils', 'underscore', 'promise'], function (utils, _) {
     DdNoise.prototype.oneShot = function (sound) {
         var duration = sound.duration;
         var context = this.context;
-        if (context.state != "running") return duration;
+        if (context.state !== "running") return duration;
         var source = context.createBufferSource();
         source.buffer = sound;
         source.connect(this.gain);
@@ -66,7 +66,7 @@ define(['./utils', 'underscore', 'promise'], function (utils, _) {
     DdNoise.prototype.play = function (sound, loop) {
         var self = this;
         var context = self.context;
-        if (context.state != "running") return null;
+        if (context.state !== "running") return null;
         return new Promise(function (resolve, reject) {
             var source = context.createBufferSource();
             source.loop = !!loop;
--- a/ddnoise.js
+++ b/ddnoise.js
@@ -64,11 +64,10 @@ define(['./utils', 'underscore', 'promise'], function (utils, _) {
     };
 
     DdNoise.prototype.play = function (sound, loop) {
+        if (this.context.state !== "running") return Promise.reject();
         var self = this;
-        var context = self.context;
-        if (context.state !== "running") return null;
         return new Promise(function (resolve, reject) {
-            var source = context.createBufferSource();
+            var source = self.context.createBufferSource();
             source.loop = !!loop;
             source.buffer = sound;
             source.connect(self.gain);
@@ -88,14 +87,12 @@ define(['./utils', 'underscore', 'promise'], function (utils, _) {
         if (this.state === SPINNING || this.state === SPIN_UP) return;
         this.state = SPIN_UP;
         var self = this;
-        var promise = self.play(self.sounds.motorOn);
-        if (promise === null) return;
-        promise.then(function () {
+        this.play(this.sounds.motorOn).then(function () {
             self.play(self.sounds.motor, true).then(function (source) {
                 self.motor = source;
                 self.state = SPINNING;
             });
-        });
+        }, function () {});
     };
 
     DdNoise.prototype.spinDown = function () {
--- a/ddnoise.js
+++ b/ddnoise.js
@@ -53,17 +53,22 @@ define(['./utils', 'underscore', 'promise'], function (utils, _) {
     };
 
     DdNoise.prototype.oneShot = function (sound) {
-        var source = this.context.createBufferSource();
+        var duration = sound.duration;
+        var context = this.context;
+        if (context.state != "running") return duration;
+        var source = context.createBufferSource();
         source.buffer = sound;
         source.connect(this.gain);
         source.start();
-        return sound.duration;
+        return duration;
     };
 
     DdNoise.prototype.play = function (sound, loop) {
         var self = this;
+        var context = self.context;
+        if (context.state != "running") return null;
         return new Promise(function (resolve, reject) {
-            var source = self.context.createBufferSource();
+            var source = context.createBufferSource();
             source.loop = !!loop;
             source.buffer = sound;
             source.connect(self.gain);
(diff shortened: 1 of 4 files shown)


ddnoise.js
```diff
@@ -53,17 +53,22 @@ define(['./utils', 'underscore', 'promise'], function (utils, _) {
...
         source.connect(this.gain);
         source.start();
-        return sound.duration;
+        return duration;
     };
 
     DdNoise.prototype.play = function (sound, loop) {
         var self = this;
+        var context = self.context;
+        if (context.state !== "running") return null;
```
mattgodbolt: It might clean the code up to return a resolved promise here (`return Promise.resolve();`). That would get rid of the `if` check for null on line 91/92.

ddnoise.js
```diff
@@ -53,17 +53,22 @@ define(['./utils', 'underscore', 'promise'], function (utils, _) {
...
         return new Promise(function (resolve, reject) {
-            var source = self.context.createBufferSource();
+            var source = context.createBufferSource();
```
mattgodbolt: Where is `context` coming from now? I can't see that there's a closure-captured `context` anywhere - does this work? :) I probably misunderstand Javascript binding rules (As usual...)

One day I'll ES6-i-fy all this... I've only been using ES6 recently and all this `var self` and `prototype.blah` is so painful to read :)

debug.js
```diff
@@ -39,7 +39,6 @@ define(['jquery', 'underscore', './utils'], function ($, _, utils) {
         };
 
         enable(false);
-        $('.initially-hidden').removeClass('initially-hidden');
```
mattgodbolt, line 42: Thanks for moving this; it was a pretty bad place to put it (sorry!)

main.js
```diff
@@ -186,11 +186,28 @@ require(['jquery', 'underscore', 'utils', 'video', 'soundchip', 'ddnoise', 'debu
         if (parsedQuery.fakeVideo !== undefined)
             video = new Video.FakeVideo();
 
+        // Recent browsers, particularly Safari and Chrome, require a user
```
mattgodbolt, line 189: Thanks for putting the comment

index.html
```diff
@@ -99,6 +99,10 @@
     </div>
 </div>
 
+<div id="audio-warning" class="alert alert-warning initially-hidden">
```
mattgodbolt, line 102: Thanks so much for making this UI experience so good!

main.js
```diff
@@ -202,6 +219,12 @@ require(['jquery', 'underscore', 'utils', 'video', 'soundchip', 'ddnoise', 'debu
             };
             soundChip.jsAudioNode.connect(audioContext.destination);
             if (!noSeek) ddNoise = new DdNoise.DdNoise(audioContext);
+
+            $audioWarningNode.toggle(false);
+            // Firefox will report that audio is suspended even when it will
```
mattgodbolt, line 224: Wow: great find. Thanks again

ddnoise.js
scarybeasts: Yes, that was a bit ham-fisted. Updated patch uses Promise.reject() which I think it conceptually what is going on here. (Promise.resolve() seemed to work but I think it may have left the ddnoise object in an inconsistent state.)

ddnoise.js
scarybeasts: I tweaked it a little for clarity. It was working due to the inner function having access to the outer function's local variables. Same trick as the "var self = this" idiom but not as clear.

ddnoise.js
```diff
@@ -64,11 +64,10 @@ define(['./utils', 'underscore', 'promise'], function (utils, _) {
     };
 
     DdNoise.prototype.play = function (sound, loop) {
+        if (this.context.state !== "running") return Promise.reject();
         var self = this;
-        var context = self.context;
-        if (context.state !== "running") return null;
         return new Promise(function (resolve, reject) {
-            var source = context.createBufferSource();
+            var source = self.context.createBufferSource();
```
mattgodbolt, line 70: Oh!!! I was purely being blind before: feel free to put it back as it was! Sorry :/

ddnoise.js
```diff
@@ -88,14 +87,12 @@ define(['./utils', 'underscore', 'promise'], function (utils, _) {
...
-        var promise = self.play(self.sounds.motorOn);
-        if (promise === null) return;
-        promise.then(function () {
+        this.play(this.sounds.motorOn).then(function () {
             self.play(self.sounds.motor, true).then(function (source) {
                 self.motor = source;
                 self.state = SPINNING;
             });
-        });
+        }, function () {});
```
mattgodbolt, line 95: I'm not sure what the second function is here for? Are you looking to catch exceptions? (I think that needs a `.catch()...` ) Also it's early in the morning here so I'm probably missing something again hahah!

Thanks for tolerating my inane comments..

ddnoise.js
scarybeasts: I think I'll keep the latest -- in terms of clarity it feels like the best of both worlds. Now, the inner function uses self exclusively for outer function access, and the outer function uses this exclusively.

ddnoise.js
scarybeasts: It's interesting, .catch() also works to resolve the problem in a quick test.

What is happening here is that when you reject a promise and the then() call doesn't have a 2nd argument (the rejection function), the Chrome Javascript console registers an exception in an sync context. I think it's harmless but it seems like a wart we could do without.

I think I prefer to keep the solution that provides an empty rejection function. This is because what we're trying to do here is handle a clean rejection by just moving along, rather than catch any exceptions. This way, we'll also get clear exception details in the console if something _does_ start excepting at a later date.

scarybeasts: That should be "registers an exception in an _async_ context"

scarybeasts: I'll abuse the approval to merge this as-is for now -- if you prefer catch() I'm happy to update to it in a subsequent PR.

mattgodbolt: Oh totally! No worries: this is just great as it is! :) You have full commit rights, I appreciate the PRs but I trust you to just make changes if that makes sense too :)

PR #215, opened 2019-08-17
Improve modeling of BRK / IRQ / NMI.

scarybeasts:
I'm not sure of any software adversely affected by the modeling inaccuracies, but in terms of things specifically fixed:

- BRK is actually a BRK #imm instruction, it reads a second byte at pc+1. In unusual circumstances, that read would have side effects if it's device I/O.
- On 6502, BRK can be "swallowed" by an NMI.
- The current code executes both an IRQ and an NMI back to back if both hit at the "same" time. During disc loading, a simultaneous IRQ and NMI is pretty common (you can add an alert() and it fires copiously). This is visible to software: the NMI will execute 7 cycles later than it should, and the NMI will have the IRQ's stack pushes visible on the stack.

This diff also reduces code duplication.

### DIFF
--- a/6502.js
+++ b/6502.js
@@ -60,6 +60,7 @@ define(['./utils', './6502.opcodes', './via', './acia', './serial', './tube', '.
 
             cpu.checkInt = function () {
                 cpu.takeInt = !!(cpu.interrupt && !cpu.p.i);
+                cpu.takeInt |= cpu.nmi;
             };
 
             cpu.setzn = function (v) {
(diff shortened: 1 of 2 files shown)


mattgodbolt:
Wow! I *never* knew BRK took the next opcode :)

6502.js
```diff
@@ -83,16 +84,43 @@ define(['./utils', './6502.opcodes', './via', './acia', './serial', './tube', '.
...
             };
 
-            cpu.brk = function () {
-                var nextByte = cpu.pc + 1;
-                cpu.push(nextByte >>> 8);
-                cpu.push(nextByte & 0xff);
-                cpu.push(cpu.p.asByte());
-                cpu.pc = cpu.readmem(0xfffe) | (cpu.readmem(0xffff) << 8);
+            cpu.brk = function (isIrq) {
+                // Behavior here generally discovered via Visual 6502 analysis.
```
mattgodbolt, line 88: Amazing! Thanks for this comment :)

6502.js
```diff
@@ -346,28 +374,7 @@ define(['./utils', './6502.opcodes', './via', './acia', './serial', './tube', '.
                     var opcode = this.readmem(this.pc);
                     this.incpc();
                     this.runner.run(opcode);
-                    if (this.takeInt) {
```
mattgodbolt, line 349: love this removal and dedupe :)

6502.opcodes.js
```diff
@@ -194,7 +194,7 @@ define(['./utils'], function (utils) {
             case "NOP":
                 return {op: "", read: arg !== undefined};
             case "BRK":
-                return {op: "cpu.brk();", extra: 6};
+                return {op: "cpu.brk(false);"};
```
mattgodbolt, line 197: Yay! :)

scarybeasts:
I should add an extra note regarding 65c12 BRK for posterity. Whereas 6502 BRK is well documented (and can be explored via Visual 6502), 65c12 BRK isn't as well documented. Accordingly, I'm not 100% sure it's correct but here's a useful reference or two:

http://wilsonminesco.com/NMOS-CMOSdif/
"A BRK instruction that was being fetched when an interrupt hit was ignored on NMOS.  (CMOS executes the BRK, then takes the interrupt.)"

http://westerndesigncenter.com/wdc/documentation/w65c02s.pdf
"BRK is executed, and then interrupt is executed."

The open question is whether 65c12 BRK contains an interrupt poll point or not. One would be needed for the IRQ to execute immediately after the BRK, so in this patch, a normally-positioned IRQ poll point was implemented for 65c12 BRK.

scarybeasts:
Re-reading what I just wrote, I'll add a TODO comment to check the 65c12 BRK poll point.

PR #219, opened 2019-08-19
Fix various bugs in the Google Drive support.

### LINKED ISSUE #113: Work around Chrome bug (was Google Drive write access is broken)
Seems the scope I've picked allows only read-only access: https://developers.google.com/drive/v2/web/scopes#google_drive_scopes 

It's not clear what the "right" setting is: the old setting (of just drive) allows read/write to all files. I just want read/write to files _I_ create...

scarybeasts:
Fixes #113 

- Using the gd: pseudo-scheme for any disc, such as disc2, resulted in the URL being modified and "corrupting" the disc parameter.
- Various automatic URL modifications would get confused between disc and disc1, often leaving conflicting parameters even those these are aliases.
- Google Drive discs were being created with size 100KB, leading to corruption when filling the SSD above 100KB (capacity is 200KB).
- Google Drive discs didn't support DSD images.
- The use of Function.prototype.apply() on a large Uint8Array was blowing out the stack, at least on Chrome.

### DIFF
--- a/fdc.js
+++ b/fdc.js
@@ -31,9 +31,7 @@ define(['./utils'], function (utils) {
         if (typeof stringData !== "string") {
             data = stringData;
         } else {
-            var len = stringData.length;
-            data = new Uint8Array(len);
-            for (var i = 0; i < len; ++i) data[i] = stringData.charCodeAt(i) & 0xff;
+            data = utils.stringToUint8Array(stringData);
         }
         var prevData = new Uint8Array(data);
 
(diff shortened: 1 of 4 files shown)


scarybeasts:
@mattgodbolt I'll fix the lint errors tomorrow, but I think it's ready for initial review.

fdc.js
```diff
@@ -61,12 +61,11 @@ define(['./utils'], function (utils) {
         var data;
         var i;
         var dataString = localStorage[discName];
-        var isDsd = false;
-        if (name.endsWith(".dsd")) isDsd = true;
+        var nameDetails = utils.discImageSize(name);
```
mattgodbolt, line 62: thanks for extracting this

fdc.js
```diff
@@ -77,8 +76,7 @@ define(['./utils'], function (utils) {
             for (i = 0; i < len; ++i) data[i] = dataString.charCodeAt(i) & 0xff;
         }
         return new BaseDisc(fdc, isDsd, data, function () {
-            var str = "";
-            for (var i = 0; i < data.length; ++i) str += String.fromCharCode(data[i]);
+            var str = utils.uint8ArrayToString(data);
```
mattgodbolt, line 76: and this...!

main.js
```diff
@@ -472,14 +478,13 @@ require(['jquery', 'underscore', 'utils', 'video', 'soundchip', 'ddnoise', 'debu
 
         function discSthClick(item) {
             utils.noteEvent('sth', 'click', item);
-            parsedQuery.disc = "sth:" + item;
-            updateUrl();
+            setDisc1Image("sth:" + item);
```
mattgodbolt, line 481: :heart: thanks, this is so much better.

scarybeasts:
Lint fixes and also:

- Upsize old 100KB images to 200KB or 400KB as appropriate to load them and prevent corruption.
- Make some more helper functions to avoid duplicated code.
- Disc name should be 8 bytes, not 12.

PR #224, opened 2019-08-24
Always clear video buffer and only render visible parts.

### LINKED ISSUE #119: CRTC fidelity - widescreen modes
Try this:
MODE 3
VDU 19,0,4,0,0,0
*HELP
VDU 23;9,7;0;0;0
VDU 19,0,1,0,0,0
When we collapse the venetian blinds, we get some stray graphics at the bottom of the screen, which don't change colour with the main part. I'm not sure if the stuff at the bottom should be there, but if it should, I think it should change colour. (I have not tested on real hardware.)

See http://stardot.org.uk/forums/viewtopic.php?p=144143#p144143

scarybeasts:
Fixes #119 

Also, the code simplification speeds up jsbeeb overall by about 5% in a default-sized MODE2. I'm getting about +0.3Mhz, 6.0Mhz -> 6.3Mhz. I'd expect the speedup to be more in something like Exile where the visible display is smaller.

### DIFF
--- a/video.js
+++ b/video.js
@@ -76,6 +76,27 @@ define(['./teletext', './utils'], function (Teletext, utils) {
             );
         };
 
+        this.clearPaintBuffer = function() {
+            var fb32 = this.fb32;
+            if (this.interlacedSyncAndVideo || !this.doubledScanlines) {
+                var line = this.oddFrame ? 1 : 0;
+                while (line < 625) {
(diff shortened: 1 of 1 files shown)


mattgodbolt:
Re: `> 2` yeah...I think so.

Gonna merge this and the drop-file thingy, and push to bbc.godbolt.org

PR #226, opened 2019-08-26
Fullscreen just the canvas and use browser fullscreen API.

### LINKED ISSUE #49: Support a fullscreen mode
Via @murkle : "How about a nice full-screen mode on F11 with no "cub" and scaled up?"

scarybeasts:
Fixes #49 

This is a re-implementation that is simpler and closer to what punters seem to want in a fullscreen (immersive!). Browsers appear competent at retaining correct aspect ratio without any extra code.

Previous behavior still available via the browsers' menus (e.g. Firefox and Chrome have a fullscreen button in the main content drop down menu).

### DIFF
--- a/main.js
+++ b/main.js
@@ -433,6 +433,10 @@ require(['jquery', 'underscore', 'utils', 'video', 'soundchip', 'ddnoise', 'debu
             if (processor.sysvia) processor.sysvia.clearKeys();
         });
 
+        $('#fs').click(function () {
+            $screen[0].requestFullscreen();
+        });
+
         document.onkeydown = keyDown;
         document.onkeypress = keyPress;
         document.onkeyup = keyUp;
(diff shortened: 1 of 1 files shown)


main.js
```diff
@@ -1292,49 +1296,6 @@ require(['jquery', 'underscore', 'utils', 'video', 'soundchip', 'ddnoise', 'debu
...
-            var cubOrigWidth = $cubMonitor.width();
-            var cubToScreenWidthRatio = $screen.width() / cubOrigWidth;
-            var navbarHeight = $("#header-bar").height();
-            const desiredAspectRatio = cubOrigWidth / cubOrigHeight;
-            const minWidth = cubOrigWidth / 4;
-            const minHeight = cubOrigHeight / 4;
-            const borderReservedSize = 100;
-            const bottomReservedSize = 100;
-
-            function resizeTv() {
```
mattgodbolt, line 1309: Losing this functionality means the screen area doesn't scale to fit the window in non-fullscreen mode. That was the main reason for most of this code: it means on my silly giant monitor I don't have a vast open black space with a small BBC micro screen in it :)  (this from my smaller laptop screen...
Screenshot from 2019-08-26 07-55-14 - normally looks like this: 
Screenshot from 2019-08-26 07-57-17 )

PR #232, opened 2019-08-27
Fix an issue where the cursor render would trigger outside the border.

### LINKED ISSUE #95: Fix video emulation glitches
Weird stripey lines in top and bottom border when scrolling the screen. Assuming a video emulation bug (might also improve Exile if fixed).  Example code fragment to reproduce:

`MODE1:REPEAT PRINT"THE QUICK BROWN FOX JUMPS OVER THE LAZY":UNTILFALSE`

scarybeasts:
Fixes #95 

Don't trigger start of cursor render if we're outside the border.
Original behavior of ULA rendering across a border is preserved (I think Rich mentioned this is real?).
Rearrange nearby code to minimize checks and tests on the common cases.
Removed an unused function.
Remove "cursorOff" variable which I don't think is needed; the logic becomes clearer.

### DIFF
--- a/video.js
+++ b/video.js
@@ -48,7 +48,7 @@ define(['./teletext', './utils'], function (Teletext, utils) {
         this.drawHalfScanline = false;
         this.oddFrame = false;
         this.teletext = new Teletext();
-        this.cursorOn = this.cursorOff = this.cursorOnThisFrame = false;
+        this.cursorOn = this.cursorOnThisFrame = false;
         this.cursorDrawIndex = 0;
         this.cursorPos = 0;
         this.interlacedSyncAndVideo = false;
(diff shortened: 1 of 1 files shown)


video.js
```diff
@@ -48,7 +48,7 @@ define(['./teletext', './utils'], function (Teletext, utils) {
         this.drawHalfScanline = false;
         this.oddFrame = false;
         this.teletext = new Teletext();
-        this.cursorOn = this.cursorOff = this.cursorOnThisFrame = false;
+        this.cursorOn = this.cursorOnThisFrame = false;
```
mattgodbolt, line 51: yay much clearer too

video.js
```diff
@@ -200,14 +200,6 @@ define(['./teletext', './utils'], function (Teletext, utils) {
             }
         };
 
-        this.clearFb = function (destOffset, numPixels) {
```
mattgodbolt, line 203: is this function now unused? yay! Thanks for tidying up

video.js
```diff
@@ -505,7 +496,7 @@ define(['./teletext', './utils'], function (Teletext, utils) {
                         break;
                     case 14:
                     case 15:
-                        this.video.cursorPos = this.video.regs[15] | (this.video.regs[14] << 8);
+                        this.video.cursorPos = (this.video.regs[15] | (this.video.regs[14] << 8)) & 0x3fff;
```
mattgodbolt, line 499: Neat

PR #233, opened 2019-08-28
Tweak rendering; turn off alpha in WebGL canvas.

scarybeasts:
- Alpha is unused and turning off can enable the browser to do less painting.
- Remove a couple of unused CSS styles.
- Be more explicit about other WebGL context parameters.

### DIFF
--- a/jsbeeb.css
+++ b/jsbeeb.css
@@ -14,11 +14,6 @@ body {
     height: 648px;
 }
 
-#cub-monitor.fullscreen {
-    width: 100%;
-    height: 100%;
-}
-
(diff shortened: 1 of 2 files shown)


canvas.js
```diff
@@ -25,7 +25,11 @@ define(['webgl-debug'], function (webglDebug) {
     };
 
     function GlCanvas(canvas) {
-        var glAttrs = {depth: false, stencil: false};
+        var glAttrs = { alpha: false,
```
mattgodbolt, line 28: Wow, today I learn! Thanks !

PR #244, opened 2019-09-02
Also attempt to run in background when the disc motor is on.

scarybeasts:
Found myself wanting this (Planet Nubium loads slow!) just the other day.
It doesn't actually work (at least in Chrome) but I think we might as well land it to register intent?

My guess is that the tape loader plays continuous noise, which is a noted condition for granting a background tab running rights. Maybe the disc motor spinning is too quiet, or the wrong type of WebAudio node? It's all an extreme mystery.


scarybeasts:
@mattgodbolt Thoughts on this one?

mattgodbolt:
Been doing some research: https://developers.google.com/web/updates/2017/03/background_tabs comes up (still looking)

mattgodbolt:
(like you said: "Applications playing audio are considered foreground and aren’t throttled." )

mattgodbolt:
Just tried hacking "speedy" to `true` if the document is hidden, but...no luck. I dunno if there's an easyway to solve this. Either the disc is too quiet like you say (though the "making sound" icon is up on the tab) or else something about the way the audio is generated in the disc system isn't quite the same...

scarybeasts:
Yeah it's a murky area. The only sane idea I have at the moment is that maybe non-zero bytes to a JS audio node callback is what is needed??

mattgodbolt:
I finally worked this out: ```
    if (document.visibilityState === "hidden") {
        wasPreviouslyRunning = running;
        if (running && !processor.acia.motorOn) {
            stop(false);
        }
```

mattgodbolt:
Gonna merge an equivalent fix in the new code rather than try and resolve this! thanks @scarybeasts ...I get there in the end...

PR #245, opened 2019-09-02
Get rid of addrLine variable to simplify.

scarybeasts:
This removes a state variable and gets rid of a bunch of shifting operations, so is presumably closer to the silicon implementation. It's also clearer to have fewer variables flying around.

### DIFF
--- a/video.js
+++ b/video.js
@@ -33,7 +33,6 @@ define(['./teletext', './utils'], function (Teletext, utils) {
         this.vertCounter = 0;
         this.scanlineCounter = 0;
         this.addr = 0;
-        this.addrLine = 0;
         this.lineStartAddr = 0;
         this.nextLineStartAddr = 0;
         this.ulactrl = 0;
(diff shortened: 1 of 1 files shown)


mattgodbolt:
Thanks! Again please feel free to develop these fixes directly in `master` and just check them in! Creating a branch, making a PR, then merging it in is unnecessary for most of these changes :)

scarybeasts:
@mattgodbolt Is it a bother if I use this workflow from time to time? I quite like the server-side validations as a double check.

mattgodbolt:
Not a bother at all! no :)  Just in case you didn't know:

* You _should_ be able to run tests locally with `make test` and `make lint`. That's all that's happening on the server. If you're not on Linux, I get that it's more of a problem...
* Checking things into a branch by itself will do the checks server-side without needing to make a PR - or should do. If this doesn't turn out to be the case I should be able to fix travis to do so.
* If you push a branch and then want to just merge it to master, you don't have to make a PR and merge it on GitHub, you can do so directly from `git` itself. (`git checkout master` to switch back to master, then `git merge <BRANCH>` to merge your `<BRANCH>` into master. Then a `git push`)

And if you knew all that, I'm sorry! `git` is a pain ...

PR #246, opened 2019-09-03
Fix Twisted Brain CRTC issues.

### LINKED ISSUE #169: Difference with CRTC vs real hardware
I am currently developing a (potential) new demo that involves lots of CRTC abuse. :)
[twister.zip](https://github.com/mattgodbolt/jsbeeb/files/1896453/twister.zip) Runs on Master 128 only.

This uses the manual vsync approach that I thought I obtaiined from retrosoftware forums BITD but can't find the original reference anywhere. (My notes say maybe Tom Seddon & Tricky but might have been RTW?)

```
	{
		lda #2
		.vsync1
		bit &FE4D
		beq vsync1 \ wait for vsync
		\now we're within 10 cycles of vsync having hit


		\delay just less than one frame
		.syncloop
		sta &FE4D \ 4(stretched), ack vsync

		\{ this takes (5*ycount+2+4)*xcount cycles
		\x=55,y=142 -> 39902 cycles. one frame=39936
		ldx #142 \2
		.deloop
		ldy #55 \2
		.innerloop
		dey \2
		bne innerloop \3
		\ =152
		dex \ 2
		bne deloop \3
		\}

		nop:nop:nop:nop:nop:nop:nop:nop:nop \ +16
		bit &FE4D \4(stretched)
		bne syncloop \ +3
		\ 4+39902+16+4+3+3 = 39932
		\ ne means vsync has hit
		\ loop until it hasn't hit

		\now we're synced to vsync
	}
```

It then sets up the CRTC to use 1 scanline per character row and 1 character row per CRTC cycle then cycle counts to update the screen buffer address in CRTC R12/R13 every scanline. On scanline 255 it sets up the necessary registers to trigger vsync at the appropriate point and form a 312 scanline signal (at least in theory.)

However, since R12/R13 are only read on a new CRTC cycle, these are reset at the end of the frame (scanline 255) ready for scanline 0 after vsync. I've hacked the code so that the first scanline is a fixed address (&3000) and this is the result I see on jsbeeb (and b-em) but on real Master hardware the last scanline is fixed at the bottom of the screen?! This is on both a CUB monitor and a more modern LCD flatscreen via RGB SCART.

I'm confused what's happening here. It might well be user error (and a misunderstanding of PAL timings) or something else. I'm pretty sure I have interlace turned off on my machine.

Source is on GitHub [here](https://github.com/bitshifters/just-rasters) (although working versions may be lurking in branches.)

scarybeasts:
Fixes #169 

Emulates a strange quirk of the Hitachi 6845.

### DIFF
--- a/video.js
+++ b/video.js
@@ -250,10 +250,9 @@ define(['./teletext', './utils'], function (Teletext, utils) {
         };
 
         this.endOfCharacterLine = function () {
-            // Still updating screen
             this.vertCounter = (this.vertCounter + 1) & 0x7f;
 
-            // Initiate vsync
+            // Initiate vsync.
             if (this.vertCounter === this.regs[7]) {
                 this.inVSync = true;
                 this.vpulseCounter = 0;
@@ -271,7 +270,7 @@ define(['./teletext', './utils'], function (Teletext, utils) {
             this.dispEnabled |= SCANLINEDISPENABLE;
             this.cursorOn = false;
 
-            // Handle vertical displayed
+            // Handle vertical displayed.
             if (this.vertCounter === this.regs[6]) {
                 this.dispEnabled &= ~VDISPENABLE;
             }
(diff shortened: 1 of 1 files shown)


video.js
```diff
@@ -381,6 +375,17 @@ define(['./teletext', './utils'], function (Teletext, utils) {
                 // Handle HSync
                 if (this.inHSync) this.handleHSync();
 
+                // Handle latching of vertical adjust pending.
```
mattgodbolt, line 378: Thanks for the clear comment! This wasn't as scary a change as I was anticipating!

PR #249, opened 2019-09-04
Make speedy mode faster by skipping painting of most frames.

scarybeasts:
This skips painting but does not skip accuracy -- the full state machine is still run.

This roughly doubles the speed of speedy mode to about 11Mhz for Twisted Brain.


video.js
```diff
@@ -411,8 +424,10 @@ define(['./teletext', './utils'], function (Teletext, utils) {
                     this.hpulseCounter = 0;
                 }
 
+                // TODO: this will be cleaner if we rework skew to have fetch
+                // indepedent from render.
```
mattgodbolt: nitpick! typo in the `indepedent`

PR #261, opened 2019-09-23
Fix regression with cursor showing in Arcadians.

scarybeasts:


video.js
```diff
@@ -449,7 +452,7 @@ define(['./teletext', './utils'], function (Teletext, utils) {
                         }
 
                         // Check cursor start.
-                        if (this.addr === this.cursorPos && this.cursorOn && this.horizCounter < this.regs[1]) {
+                        if (this.addr === this.cursorPos && this.cursorOn && !this.cursorOff && this.horizCounter < this.regs[1]) {
```
mattgodbolt, line 455: Haha :) I wonder if a state machine would work better now. This works though (and I'm probably naïve about a state machine...)

scarybeasts: I do hope to remove the final predicate in the future because it can't possibly be how the silicon is working. The cursorOn and cursorOff variables likely do match to latch bits in the silicon.

richtw1: I imagine it just looks at the pre-skewed DISPEN. You can definitely have the cursor in the left border if display is skewed and I don't think you get a second cursor above and to the right in this case.

richtw1: And agreed that the cursorOn and cursorOff latches are probably exactly what the chip does.

PR #264, opened 2019-09-26
Emulate CTS a little better.

scarybeasts:
This includes inhibiting TDRE correctly.
Previously, ?&FE08 on a fresh boot gave 10, which is not supposed
to be a possible combination.
We now give 8, matching a real beeb.

### DIFF
--- a/acia.js
+++ b/acia.js
@@ -12,10 +12,6 @@ define([], function () {
         self.rs423Handler = rs423Handler;
         self.rs423Selected = false;
         self.motorOn = false;
-        // TODO: set clearToSend accordingly; at the moment it stays low.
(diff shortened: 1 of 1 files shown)


acia.js
```diff
@@ -54,8 +50,12 @@ define([], function () {
                 return self.dr;
             } else {
                 var result = (self.sr & 0x7f) | (self.sr & self.cr & 0x80);
-                if (!self.clearToSend) result &= ~0x02; // Mask off TDRE if not CTS
-                result = result | 0x02 | 0x08;
+                // MC6850: "A low CTS indicates that there is a Clear-to-Send
```
mattgodbolt, line 53: Thanks so much for citing the references for this stuff.

PR #266, opened 2019-09-26
End the tape carrier tone if the motor is switched off.

scarybeasts:


acia.js
```diff
@@ -39,11 +39,12 @@ define([], function () {
         };
 
         self.setMotor = function (on) {
-            if (on && !self.motorOn)
+            if (on && !self.motorOn) {
                 runTape();
-            else {
+            } else if (!on && self.motorOn) {
```
mattgodbolt, line 44: thanks for putting `{`s in. I was very lazy here :/

PR #267, opened 2019-09-26
Fix emulation of DCD flag in 6850 status register.

scarybeasts:


acia.js
```diff
@@ -62,6 +65,16 @@ define([], function () {
                 if (result & 0x08) {
                     result &= ~0x02;
                 }
+
+                // MC6850: "It remains high after the DCD input is returned low
+                // until cleared by first reading the Status Register and then
+                // the Data Register".
+                // Testing on a real machine shows that only the Data Register
```
mattgodbolt, line 72: "testing on a real machine" :heart:

PR #275, opened 2019-10-04
Count CRTC frames at end of vertical display.

scarybeasts:
Previously this was done at vsync and testing shows that was not
correct.

### DIFF
--- a/video.js
+++ b/video.js
@@ -22,8 +22,7 @@ define(['./teletext', './utils'], function (Teletext, utils) {
         this.bitmapY = 0;
         this.renderY = 0;
         this.oddClock = false;
-        this.frameCountCrtc = 0;
-        this.frameCountVsync = 0;
+        this.frameCount = 0;
         this.inHSync = false;
         this.inVSync = false;
         this.hadVSyncThisRow = false;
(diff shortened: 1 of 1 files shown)


scarybeasts:
For posterity, a test is:

MODE2:?&FE00=6:?&FE01=100

Correct behavior, on a Hitachi CRTC, is cursor isn't blinking.

video.js
```diff
@@ -419,8 +421,11 @@ define(['./teletext', './utils'], function (Teletext, utils) {
                 // The 6845 in the BBC will notice this equality on any clock
                 // tick, not just at the end of a character row.
                 // Required by Wave Runner.
-                if (this.vertCounter === this.regs[6]) {
+                if (this.vertCounter === this.regs[6] && (this.dispEnabled & VDISPENABLE)) {
                     this.dispEnabled &= ~VDISPENABLE;
+                    // Perhaps surprisingly, this happens here. Both cursor
+                    // blink and interlace cease if R6 > R4.
+                    this.frameCount++;
```
mattgodbolt, line 428: Just checking...this will wedge after (2**53 / 50) seconds... that's 5,712,328 years. We should be ok :D (but I had to check...the cpucycles counter used to suffer and that _did_ eventually break...)

scarybeasts:
It probably will (insofar as cursor blinking and interlace stopping).

The hardware implementation is probably just an "& 15" but letting it run for logging might look nicer. Also the frame skip logic uses this counter. We could make it & 2^32-1 perhaps?

mattgodbolt:
I think we're ok. 5 billion years is a long time to run an emulator...

PR #276, opened 2019-10-05
Fix various vertical adjust bugs and quirks.

scarybeasts:
- Advance row address while in vertical adjust.
- Fix regression with missing first vertical adjust line.

### DIFF
--- a/video.js
+++ b/video.js
@@ -268,12 +268,13 @@ define(['./teletext', './utils'], function (Teletext, utils) {
         this.endOfFrame = function () {
             this.vertCounter = 0;
             this.nextLineStartAddr = (this.regs[13] | (this.regs[12] << 8)) & 0x3FFF;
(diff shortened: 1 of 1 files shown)


video.js
```diff
@@ -282,50 +283,54 @@ define(['./teletext', './utils'], function (Teletext, utils) {
...
 
         this.endOfScanline = function () {
             if (this.scanlineCounter === this.regs[11]) this.cursorOff = true;
 
             this.vpulseCounter = (this.vpulseCounter + 1) & 0x0F;
 
-            if (this.vertAdjustPending) {
-                this.scanlineCounter = 0;
-                this.inVertAdjust = true;
+            var r9Hit = (this.scanlineCounter === this.regs[9]);
```
mattgodbolt, line 293: It's a good sign that the code's getting simpler and being phrased more like the control signals we suspect are inside the chip!

PR #278, opened 2019-10-08
Fix interlace timing and modeling.

scarybeasts:
- Odd and even interlace frames are now the same length.
- The even frame dummy raster is a vertical adjust scanline.

### DIFF
--- a/video.js
+++ b/video.js
@@ -20,7 +20,6 @@ define(['./teletext', './utils'], function (Teletext, utils) {
         this.regs = new Uint8Array(32);
         this.bitmapX = 0;
         this.bitmapY = 0;
-        this.renderY = 0;
         this.oddClock = false;
         this.frameCount = 0;
         this.inHSync = false;
@@ -47,7 +46,6 @@ define(['./teletext', './utils'], function (Teletext, utils) {
         this.displayEnableSkew = 0;
         this.ulaPal = utils.makeFast32(new Uint32Array(16));
         this.actualPal = new Uint8Array(16);
-        this.drawHalfScanline = false;
         this.teletext = new Teletext();
         this.cursorOn = false;
         this.cursorOff = false;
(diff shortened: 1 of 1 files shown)


video.js
```diff
@@ -362,26 +365,20 @@ define(['./teletext', './utils'], function (Teletext, utils) {
...
                 }
 
-                this.bitmapY++;
+                // The CRT vertical beam speed is constant, so this is actually
+                // an approximation that works if hsyncs are spaced evenly.
+                this.bitmapY += 2;
+
                 // If no VSync occurs this frame, go back to the top and force a repaint
-                if (this.bitmapY >= 384) {
+                if (this.bitmapY >= 768) {
```
mattgodbolt, line 373: I've been suspicious of this constant a good long while: what was the rationale for bumping it? (did you find a way to expose this?)

Thanks again for your tireless work on this!

video.js
```diff
@@ -362,26 +365,20 @@ define(['./teletext', './utils'], function (Teletext, utils) {
                     this.bitmapX -= 4;
                 }
 
-                this.bitmapY++;
+                // The CRT vertical beam speed is constant, so this is actually
+                // an approximation that works if hsyncs are spaced evenly.
+                this.bitmapY += 2;
```
mattgodbolt, line 370: Makes perfect sense!

video.js
scarybeasts: This constant in some sense hasn't "changed". I removed "this.renderY" in order to make the code simpler and less fragile to change. Now there's only the one variable, "this.bitmapY" for tracking where the CRT vertical beam position is within the canvas.
As part of that change, "this.bitmapY" doubled in value, including this constant.
The constant now does represent PAL lines, so IMHO it's a bit more easy to see that it's a little more than ~624, the standard number of PAL lines per PAL frame.

Unfortunately I have no idea if there's a better value for the constant!

mattgodbolt: Thanks fella, got it :)

PR #282, opened 2019-10-17
Add pause and step feature.

kieranhj:
Add pauseEmu state, similar to debugger but no UI.
Add stepEmuWhenPaused state to continue and display a single frame of the emulator.

Pause on CTRL-END. 'g' to go/continue (same as debugger), 'n' to step to next frame (similar to debugger next instruction).

Not sure if you approve of my single-step nonsense but I find it useful!

### DIFF
--- a/main.js
+++ b/main.js
@@ -42,6 +42,8 @@ require(['jquery', 'underscore', 'utils', 'video', 'soundchip', 'ddnoise', 'debu
         var fastAsPossible = false;
         var fastTape = false;
         var noSeek = false;
+        var pauseEmu = false;
+        var stepEmuWhenPaused = false;
         var audioFilterFreq = 7000;
         var audioFilterQ = 5;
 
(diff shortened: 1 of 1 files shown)


main.js
```diff
@@ -1323,6 +1340,10 @@ require(['jquery', 'underscore', 'utils', 'video', 'soundchip', 'ddnoise', 'debu
                     dbgr.debug(processor.pc);
                     throw e;
                 }
+                if (stepEmuWhenPaused) {
```
mattgodbolt, line 1343: All looks good, maybe a better name though? Naming is hard...I know...but "stop emu when paused" confuses me: what is pausing? what's a step? Would 'singleStepFrame' make more sense? We can probably make a single keypress (shift G maybe?) in the regular debugger do this, instead of having another mode: but this works too :) I want jsbeeb to be useful to your debugging style, and I'm not 100% sure of the use case :)

mattgodbolt: Ahh (leaving original comment in place!)

I see: pausing is a new state that doesn't drop into the debugger!

Looks good to me !

PR #285, opened 2019-10-21
Fix bugs causing the disc spinning noise to never stop.

scarybeasts:
- Hard reset would lose track of state.
- The 1770 driver would often cause bad state upon an error.


scarybeasts:
This was super annoying!

mattgodbolt:
Oh cripes! Thanks mate!

PR #295, opened 2019-11-13
Fix CRT beam dot position in debug mode.

### LINKED ISSUE #291: fake CRTC dot out of step with actual location
Now the screen doesn't draw below the point where the CRTC dot is... the crappy dot I draw for the electron beam is clearly in the wrong place... #274 
Screenshot from 2019-10-28 17-50-41

It's roughly the right spot left to right, but far too far down the screen. One solution would be to just kill this feature :)

scarybeasts:
Fixes #291

### DIFF
--- a/video.js
+++ b/video.js
@@ -131,11 +131,7 @@ define(['./teletext', './utils'], function (Teletext, utils) {
         this.debugOffset = function (x, y) {
             if (x < 0 || x >= 1024) return -1;
             if (y < 0 || y >= 768) return -1;
-            var renderY = (y << 1);
-            if (this.interlacedSyncAndVideo || !this.doubledScanlines) {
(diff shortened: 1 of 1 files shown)


video.js
```diff
@@ -131,11 +131,7 @@ define(['./teletext', './utils'], function (Teletext, utils) {
         this.debugOffset = function (x, y) {
             if (x < 0 || x >= 1024) return -1;
             if (y < 0 || y >= 768) return -1;
-            var renderY = (y << 1);
-            if (this.interlacedSyncAndVideo || !this.doubledScanlines) {
-                this.renderY += (this.frameCount & 1);
-            }
-            return renderY * 1024 + x;
+            return (y * 1024) + x;
```
mattgodbolt, line 134: thanks :)

PR #314, opened 2020-12-31
Allow tokenising longer input lines

ojwb:
The code in the BASIC ROM we use copies down the tail of the line
after substituting a token with a loop indexed on Y, which limits
the untokenised tail to 255 bytes (plus a terminating CR).

Intercept this copy loop and perform the copying in JS to allow
longer inputs to be tokenised.  The output still has to fit in the BASIC
line length limit.

I would flag this as WIP but I can't seem to see how to in github.  Still to resolve is that I've attempted to add a test that an overlong input gives a suitable exception, but the testharness seems to complain this takes more than 2 seconds.  I think I'm out of the javascript depth here so pointers appreciated.

### DIFF
--- a/basic-tokenise.js
+++ b/basic-tokenise.js
@@ -19,13 +19,34 @@ define(['utils', 'models', 'fake6502', 'promise'],
                 }
                 cpu.writemem(offset + line.length, 0x0d);
                 var safety = 20 * 1000 * 1000;
+                var result = "";
                 while (cpu.s <= 0xf0) {
                     cpu.execute(1);
                     if (--safety === 0) {
                         break;
                     }
+                    if (cpu.pc === 0x8ea1) {
+                        // Intercept the subroutine in the BASIC ROM which replaces a keyword
+                        // with a token and copies down the tail of the line.  The 6502 code
+                        // uses the Y register to index the copy and fails if the untokenised
+                        // tail is longer than 255 bytes.  It also makes tokenisation O(n²)
+                        // for a line with a lot of tokens.
+                        //
+                        // Instead we copy out the newly processed part and advance the pointer
+                        // in 0x37/0x38 to the unprocessed part.
+                        let to = cpu.readmemZpStack(0x38) << 8 | cpu.readmemZpStack(0x37);
+                        while (offset < to) {
+                            result += String.fromCharCode(cpu.readmem(offset));
+                            offset++;
+                        }
+                        result += String.fromCharCode(cpu.a);
+                        offset += cpu.y;
+                        cpu.writememZpStack(0x37, offset & 0xff);
+                        cpu.writememZpStack(0x38, (offset >>> 8) & 0xff);
+                        ++offset;
+                        cpu.pc = 0x8ea4;
+                    }
                 }
-                var result = "";
                 for (i = offset; cpu.readmem(i) !== 0x0d; ++i) {
                     result += String.fromCharCode(cpu.readmem(i));
                 }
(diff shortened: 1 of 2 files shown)


mattgodbolt:
Re: WIP - yeah I dunno why "Draft PR"s have disappeared...I see them on my (corporate) servers...

For now I'll make a "do not merge" label. LMK when this is ready for a look

ojwb:
I was hoping you'd take a quick look and help me solve the new testcase, but I've managed to do that by myself (though it may be a terrible solution as I don't really know much about async JS...)

I've also reworked to intercept and replace the whole subroutine to avoid the O(n²) behaviour of the BASIC tokenising code which avoids the testcase taking as long, but I don't really get how a test harness can just fail a test based just on it taking more than 2 seconds - surely that means lots of tests will fail if run on a much slower machine than the one the tests were written on?

murkle:
Does it need to work with all versions of BBC Basic (BASIC I, BASIC II, BBC Master Basic MOS 3.2/3.5, maybe others)?

I guess this is hard-coded to BASIC II
`if (cpu.pc === 0x8ea1) {`

ojwb:
> Does it need to work with all versions of BBC Basic (BASIC I, BASIC II, BBC Master Basic MOS 3.2/3.5, maybe others)?

No, this code explicitly only runs on the configuration `models.basicOnly` which is defined as:

```
    var basicOnly = new Model("Basic only", ["Basic only"], ["master/mos3.20"], false, true, masterSwram, fdc.WD1770);                                          
```

> I guess this is hard-coded to BASIC II
> `if (cpu.pc === 0x8ea1) {`

To BASIC on a BBC Master actually, but the code is already specific to that version due to the address of the tokenisation code it jumps to with:

```
    cpu.pc = 0x8db2;
```

This probably could all be generalised to allow tokenising for a particular version of BBC BASIC, but really I don't think that should be a requirement for merging this patch.

murkle:
Thanks!

I suggest adding the magic numbers into variables with some comments so it makes sense to someone reading the code in 5 years :)
```
0x8ea1
0x8db2
```

Maybe also check ?&8ea1 and ?&8db2 contain the right values for a belt-and-braces approach

ojwb:
I've documented most of the magic numbers in this code (including the stack pointer trick used to detect when we're done - that confused me initially as I wrongly assumed that `cpu.s` was the status flags.)

I'm not sure checking the contents of the locations helps much.  JSR instructions are common and really we'd need to checksum the whole area of the ROM that we make use of here to reliably detect a different ROM version. But the existing tokenisation tests are bound to fail if someone does that, so we'll already catch if that happens.

ojwb:
@mattgodbolt I think this is ready.

Except ... I just noticed I've used `writememZpStack` to write to the zero page pointer, whereas the existing code just uses `writemem`.  Do you have a preference to use one or the other consistently?  (`writememZpStack` avoids needlessly checking for paged memory, but the overhead is probably negligible so maybe always using `writemem` in code like this is preferred.)

mattgodbolt:
Hey @ojwb - yeah `writememZpStack` is probably misnamed (`writememUnpagedNonHardwareFast`?). I mean, the overhead _is_ negligible for this use case, so `writemem` is fine too, but I don't mind :)

tests/unit/test-tokenise.js
```diff
@@ -43,4 +55,10 @@ describe('Tokeniser', function () {
     it('handles REM', function (done) {
         check(done, "10REM I am a monkey", "\r\x00\x0a\x13\xf4 I am a monkey\r\xff");
     });
+    it('handles extra long input', function (done) {
+        check(done, "10" + "ENVELOPE".repeat(251), "\r\x00\x0a\xff" + "\xe2".repeat(251) + "\r\xff");
```
mattgodbolt, line 59: TIL `.repeat`!

basic-tokenise.js
```diff
@@ -5,27 +5,64 @@ define(['utils', 'models', 'fake6502', 'promise'],
...
+                // Set the paged ROM latch to page in the BASIC.
                 cpu.writemem(0xfe30, 12);
                 for (var i = 0; i < line.length; ++i) {
                     cpu.writemem(offset + i, line.charCodeAt(i));
                 }
                 cpu.writemem(offset + line.length, 0x0d);
                 var safety = 20 * 1000 * 1000;
-                while (cpu.s <= 0xf0) {
+                var result = "";
+                while (cpu.s <= stackTop) {
```
mattgodbolt, line 38: Thanks for extracting all the constants, much better.

mattgodbolt (approved):
Looks great: thanks @ojwb

PR #335, opened 2021-12-22
New audio!

mattgodbolt:
* Outputs audio at chip frequency
* Spits it out as buffers to a web audio thread that then (terribly) downsamples it.
* Various queues and stats
* Horrible hacks to try and keep things in sync.


web/audio-handler.js
```diff
@@ -1,54 +1,91 @@
...
+        await this.audioContext.audioWorklet.addModule(
+            new AudioWorklet(
+                new URL("./audio-renderer.js", import.meta.url)
+            )
+        );
         if (audioFilterFreq !== 0) {
             this.soundChip.filterNode = this.audioContext.createBiquadFilter();
             this.soundChip.filterNode.type = "lowpass";
             this.soundChip.filterNode.frequency.value = audioFilterFreq;
             this.soundChip.filterNode.Q.value = audioFilterQ;
```
8bitkick: For a cleaner separation of web API I/O vs pure emulation code,`filterNode` doesn't need to be instanced as a property of soundChip

8bitkick: (I'm only noticing this due to my weird use case running JSbeeb in a worker)

mattgodbolt: Thanks! I think I put it there purely to "hang on to it" as I think in some cases if it gets GC'd maybe it stops working?

mattgodbolt:
Webpack's lack of decent support still gives me pause, but...I brought this up to date at least.

mattgodbolt:
now we're on vite I can consider looking into this again

mattgodbolt:
now we're on vite I can consider looking into this again

web/audio-handler.js
mattgodbolt: It's held on to elsewhere...so...yay! fixed!

PR #337, opened 2022-01-21
Fix for browsers without Web Audio.

pwombwell:
FakeKeyNoise() is on KeyNoise, not DdNoise.


mattgodbolt (approved):
Oops! Great spot: thank you!

PR #348, opened 2022-05-03
Add all configured models to the config dialog

### LINKED ISSUE #347: Models not listed in UI
Is there a reason why the UI only lists BBC B, Master 128 and Master Turbo, when there are more models defined? Happy to send a PR with more models in, or even to see if we can dynamically make the list from the defined models.

codeaholics:
This resolves #347, but I realise you may have left some models off the config dialog on purpose, so reject at will!


mattgodbolt:
This is great! thanks :)

PR #350, opened 2022-05-05
Only update the URL and menubar on confirmation of model change

### LINKED ISSUE #349: Config UI: updates state at before confirmation
When interacting with the model dropdown in the config UI, the following unexpected behaviours can be observed:

- Changing the model dropdown causes the menubar to change state, even though no actual model change has happened yet
- Clicking `Close` causes the URL to also update, even if the user then declines the confirmation dialog

This leaves the UI in a state where neither the menubar nor the URL reflect the actual running model, and, if the user refreshes (or copies the URL, etc.) then they will end up with a different model.

dy-tray:
Fixes #349 

Interaction with config dialog model list now doesn't update any state outside of the dialog until the users clicks `Close` AND confirms their changes.

### DIFF
--- a/config.js
+++ b/config.js
@@ -6,7 +6,11 @@ export function Config(onClose) {
     let changed = {};
     this.model = null;
     const $configuration = document.getElementById('configuration');
-    $configuration.addEventListener('show.bs.modal', () => changed = {});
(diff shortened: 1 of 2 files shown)


dy-tray:
Please delete me!

PR #354, opened 2022-05-16
Emulation of teletext adaptor and the Music 5000

robsonjb1:
Hi, this PR is a port of the teletext adaptor and Music 5000 from BeebEm. It also includes the sample teletext data (recorded from a live stream in 2006).
Additionally, the Music 5000 emulation is also added. (This utilises audio worklets so is currently not supported yet by Safari).

I have hosted this fork here if you would like to take a look.
https://jsbeeb.azurewebsites.net

Two additional models have been added.
1/ Teletext. This includes the ATS teletext support ROM. Do a *TELETEXT to start. Shift F1-F4 to change channels.
2/ Music 5000. This includes the AMPLE ROM. Boot from the Music 5000 system disk (included in cat).

I think the work you've done on jsbeeb is outstanding. I'm not sure of the direction you wish to take it so understand if you don't feel 'bloating' out the app with specific hardware emulation is the way to go. (But for me, I very much enjoyed getting involved in the beeb scene again...!)

Jason Robson 
jason@jasonrobson.com


mattgodbolt:
Wow! Thanks!! I'll try and find some time to do a review soon!! Thanks so much

mattgodbolt (commented):
Quick things - please put the `.gitignore` in `discs` back and remove the `.DS_Store` file - if you need to add files in `discs` you can add them without removing the file (this is to stop me accidentally checking in copyrighted discs).

Also - do we have permission to put those SSDs in? With the welcome disc we were ok (as best I can tell it's PD now), and I have explicit permission for Elite.

mattgodbolt:
also you checked the files in as `SHOUTCASE` but refer to them in the code in `lowercase` - can you make it all lower case please? (flyby quick review!!)

README.md
```diff
@@ -1,5 +1,35 @@
 [![jsbeeb tests](https://github.com/mattgodbolt/jsbeeb/actions/workflows/test-and-deploy.yml/badge.svg)](https://github.com/mattgodbolt/jsbeeb/actions/workflows/test-and-deploy.yml)
 
+
+
+Music 5000 emulation
+--------------------
+
+I've ported over the Beebem code for the Music 5000. This version uses Audio Worklets so
```
mattgodbolt: Who is the "I" here? :) All this being at the top of the README is confusing - is the plan to remove this?

robsonjb1: I've undone the update to README.md

robsonjb1:
> Also - do we have permission to put those SSDs in? With the welcome disc we were ok (as best I can tell it's PD now), and I have explicit permission for Elite.

I'm looking into the copyright situation... I've updated the other initial comments (.gitignore, lowercase filenames)

dy-tray:
Amazing! Question for Matt, does it make sense for additional hardware to be defined in the `model`? Or should there be a different mechanism for attaching different hardware? What if I wanted a teletext adaptor AND a Music 5000? For the base computer with its CPU (and maybe second processor), the `model` makes sense, but looking forward, I'm not even sure that the choice of DFS ROM should be in there either.

codeaholics:
Amazing! Question for Matt: does it still make sense for peripheral hardware to be define in the `model`? What happens if I want a machine with both a teletext adaptor AND a Music 5000 installed? I think the `model` makes sense for the base machine, CPU, maybe second processor, maybe pre-installed ROMs, etc. But I wonder if peripherals and maybe things like the DFS ROM version, etc., should perhaps have a different UI to enable them to be plugged and unplugged?

mattgodbolt:
It probably shuld be more "add on" based but I punted on the UI for that kind of thing :) So for now "Models" are all we got 😁

models.js
```diff
@@ -29,7 +31,9 @@ const masterSwram = [
     false, false, false, false];
 const tube65c02 = new Model("Tube65C02", [], ["tube/6502Tube.rom"], false, false);
 export const allModels = [
-    new Model("BBC B", ["B"], ["os.rom", "BASIC.ROM", "b/DFS-0.9.rom"], true, false, beebSwram, I8271),
+    new Model("BBC B", ["B"], ["os.rom", "BASIC.ROM", "b/DFS-0.9.rom"], true, false, beebSwram, I8271, null, false, false),
+    new Model("BBC B with Teletext", ["BTeletext"], ["os.rom", "BASIC.ROM", "b/DFS-0.9.rom", "ATS-3.0.ROM"], true, false, beebSwram, I8271, null, true),
```
mattgodbolt: Looks like these need to be case-matched with the ROM files you checked in :)

```suggestion
    new Model("BBC B with Teletext", ["BTeletext"], ["os.rom", "BASIC.ROM", "b/DFS-0.9.rom", "ats-3.0.ROM"], true, false, beebSwram, I8271, null, true),
```

models.js
```diff
@@ -29,7 +31,9 @@ const masterSwram = [
...
+    new Model("BBC B with Music 5000", ["BMusic5000"], ["os.rom", "BASIC.ROM", "b/DFS-0.9.rom", "AMPLE.ROM"], true, false, beebSwram, I8271, null, true, true),
```
mattgodbolt:
```suggestion
    new Model("BBC B with Music 5000", ["BMusic5000"], ["os.rom", "BASIC.ROM", "b/DFS-0.9.rom", "ample.rom"], true, false, beebSwram, I8271, null, true, true),
```

music5000.js
```diff
@@ -0,0 +1,195 @@
...
+const FREQ_DISABLE = (1 << 0);
+const NOT_FREQ_DISABLE = 0xfe;
+
+// Channel register sets
+const REG_SET_NORMAL = 0;
+const REG_SET_ALT = 1;
+
+
+export function Music5000(onBuffer) {
+	this._onBufferMusic5000 = onBuffer;
```
mattgodbolt: Can you please use four spaces for indenting like the rest of the codebase? I really should set up automatic formatting -- sorry!

music5000.js
```diff
@@ -0,0 +1,195 @@
...
+const DATA_VALUE = (0x7f);
+const FREQ_DISABLE = (1 << 0);
+const NOT_FREQ_DISABLE = 0xfe;
+
+// Channel register sets
+const REG_SET_NORMAL = 0;
+const REG_SET_ALT = 1;
+
+
+export function Music5000(onBuffer) {
```
mattgodbolt: For what it's worth I'm now moving more things to newer JS versions - you can use `class` here instead of the old-style function-based classes. Not a problem either way, and most of the code still uses the older style, so it fits right in :)

music5000.js
```diff
@@ -0,0 +1,195 @@
...
+			console.log('Music 5000: initialisation');
+
+			// Build the D2A table
+			var i = 0;
+			for (var chord = 0; chord < 8; chord++)
+			{
+				var val = this.chordBase[chord];
+				for (var step = 0; step < 16; step++)
+				{
+					this.D2ATable[i] = parseInt(val * 4); // Multiply up to get an integer
```
mattgodbolt: When I think of `parseInt` I'm usually thinking of parsing a string - if you need to get an integer value here you can do either: `Math.ceil()` or `Math.floor` or `val | 0` (which has the side-effect of forcing `val` to be an integer).

teletext_adaptor.js
```diff
@@ -0,0 +1,188 @@
...
+   0-1      Channel select
+   2        Teletext Enable
+   3        Enable Interrupts
+   4        Enable AFC (and mystery links A)
+   5        Mystery links B
+
+*/
+
+export function TeletextAdaptor(cpu) {
+	var self = this;
```
mattgodbolt: ditto tabs :) If reformatting is'nt trivial I can reformat after merging (and I must set up `prettier` or similar to format everything...)

mattgodbolt (commented):
This is an amazing start: thank you!!

Mostly minor comments here and most can be ignored, depending on your appetite to learn JS!

The major things are:

* the `dat` files are not bundled with webpack. We'll have to add them to webpack.config
* as a a result of this when I bundle for "producition" I get `TypeError: this.music5000.reset is not a function` which I think is because of an earlier `Unable to load teletext/txt0.dat`

this is in "production" mode - things seem to work (inasmuch as I can tell) in regular development mode.

I can help you with the last parts as it's not really that obvious (and doesn't show up without building). If you want to take a look at a built thing; `npm run build` to build, then serve the files in `out/dist` (I use `python -mhttp.server` or similar), but I digress.

Sorry it took so long to get back to you: I'm excited and thank you once again!

Let me know how you feel about the changes, and I can merge in and try and fix up the webpack stuff if you'd like.

6502.js
```diff
@@ -596,7 +599,25 @@ export function Cpu6502(model, dbgr, video_, soundChip_, ddNoise_, cmos, config)
             return this.ramRomOs[this.osOffset + (addr & 0x3fff)];
         }
         addr &= 0xffff;
+            
+        if (model.hasMusic5000)
+        {
+            if (addr == 0xfcff) {
```
mattgodbolt: All the `==` and `!=` here and elsewhere need to be `===` and `!==` because...javascript is awful. `===` is "equality" but `==` is "equalish" - like `0 == "0"` is `true` but `0 === "0"` is `false`. I can fix these up (my IDE screams blue murder on seeing them hahah)

6502.js
```diff
@@ -596,7 +599,25 @@ export function Cpu6502(model, dbgr, video_, soundChip_, ddNoise_, cmos, config)
             return this.ramRomOs[this.osOffset + (addr & 0x3fff)];
         }
         addr &= 0xffff;
+            
+        if (model.hasMusic5000)
```
mattgodbolt, line 603: For all these `model.has*` checks it might be simpler and less error-prone to check `if (this.music5000)` instead . That way whether the model said it had a 5000 or not, we don't try and call methods on an object that might not be there. (ditto teletext)

mattgodbolt: ...for example if you pick a music5000 model BUT the audio worklet fails to initialise (as is happening on my machine) you don't get a crash here and in other place)

mattgodbolt:
So...I need to look into the webpack audio worklet thing. The issue I see in "release" with the `web5000.reset is not a function` is:

* You pick a music 5000 (model.has5000 = true)
* you reset
* the audio worklet fails (because of webpack things; but could also be for browser reasons on other folks' mahcines), so even though `has5000=true` the `music5000` object is `null`
* the main code checks `model.has5000` and if it's true, blindly assumes that `this.music5000` must be non-null.

I will try look into the worklet thing

models.js
robsonjb1: I've changed the case. (Genuine question: is this a stylistic thing or is there an actual functional issue in some browsers). Disclaimer: I come from the type-insensitive Windows world ;)

music5000.js
robsonjb1: This is confusing me! In my source code editor (Visual Studio Code) I am just seeing a TAB indentation here. When I look at the code this is translated to whatever my GitHub settings are (by default I think a TAB is 8 characters).

music5000.js
robsonjb1: I can switch these over to classes if you would prefer? I did it like this purely to match the rest of the code I was seeing. Happy either way...

music5000.js
robsonjb1: Changed to Math.floor ;)

teletext_adaptor.js
robsonjb1: Please see earlier comment, these appear to me to be TAB's in the source code - but just displayed by the default GitHub viewer as 8 spaces. I'm a newbie here so please let me know if I've missed something.

6502.js
robsonjb1: Changed to ===

6502.js
robsonjb1: changed to `this.music5000` etc...

models.js
mattgodbolt: It's a functional issue: only windows web servers are case insensitive :) Mac and UNIX (and AWS and CloudFront and...) are all case-sensitive, so I was getting "404 Not found" for these. :)

music5000.js
mattgodbolt: That's the issue - the rest of the codebase is indented with 4 spaces, not tabs. Can you use that please? Else I'll need to reformat it. I don't use tabs to indent, I use spaces (which means, rather rudely, I get to pick the indentation size and everyone else has to suffer it...I know that's not idea but I don't fancy mixing tabs and spaces now)

music5000.js
mattgodbolt: Whichever you fancy :) I don't want to force you through hoops; so thisi s fine as-is, but I'll (sloowwwwwly) be moving towards JS classes in the rest of the codebase.

6502.js
mattgodbolt: Thanks!

6502.js
mattgodbolt: You :star: - thanks!

mattgodbolt:
> > (And I guess I've glossed over the copyright stuff - does beebem ship the ROMs et al?)
> 
> Yes. Beebem ships with the ATS (Acorn) and AMPLE ROM's. It also has the Music 5000 system disk. I'm pretty sure this is where I got them from originally.

Then I think we're ok :) thanks for that!

mattgodbolt (approved):
Looks amazing!!! Thank you so much! Will merge shortly - the live site will update son after!

Thank you so very much for all this!

mattgodbolt:
(and thanks for choking back the spaces vs tabs thing ;) I know how contentious that is, and i don't really have a strong axe to grind but mostly "keep to what was there before" :)

mattgodbolt:
This is now live! I'll let you do any necessary announcements on `*.` :) Thanks again

robsonjb1:
Thanks Matt, pleasure helping out. The Music 5000 and Teletext stuff were fond memories from my school days so I'm really happy to see it in jsbeeb...!

robsonjb1:
Thanks Matt,
I’ll take a look at your comments in the next day or so.
You can probably tell this is the first PR I’ve ever created so you will have to bear with me :-) I’m completely new to contributing to Open Source projects on GitHub.

I did post a link to my fork on stardot which is why there are all the readme ‘I’ references.
‘I’ will sort that out as well :)

I will need to research the copyright status of the ATS ROM and the Music 5000 system disk. Both were included with Beebem so I have to admit I didn’t think about including it…

Thanks,
Jason

Sent from my iPhone

On 16 May 2022, at 21:18, Matt Godbolt ***@***.***> wrote:

﻿

@mattgodbolt commented on this pull request.

________________________________

In README.md<https://nam12.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2Fmattgodbolt%2Fjsbeeb%2Fpull%2F354%23discussion_r874121107&data=05%7C01%7C%7C5a85dda13acb4fb82bdc08da377943ff%7C84df9e7fe9f640afb435aaaaaaaaaaaa%7C1%7C0%7C637883291214766929%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C3000%7C%7C%7C&sdata=83xcUAGjrUdmTX7RlgBj9vcHxQBcrn56B1dD5NFRQPY%3D&reserved=0>:

> @@ -1,5 +1,35 @@
 [jsbeeb tests](https://github.com/mattgodbolt/jsbeeb/actions/workflows/test-and-deploy.yml)

+
+
+Music 5000 emulation
+--------------------
+
+I've ported over the Beebem code for the Music 5000. This version uses Audio Worklets so


Who is the "I" here? :) All this being at the top of the README is confusing - is the plan to remove this?

robsonjb1:
> awesome! this is all live anyway :)

Wierd! I think GitHub had a bit of a wobble, I wrote that months ago!

chrisjj:
>  Beebem ships with the ATS (Acorn) and AMPLE ROM's. It also has the Music 5000 system disk.

Really? I don't see any mention of Music 5000 support in BeebEm [here](https://github.com/stardot/beebem-windows/blob/master/README.md).

chrisn:
Have a look at https://github.com/stardot/beebem-windows/blob/master/Help/m5000.html

PR #358, opened 2022-05-23
Add .DS_Store to .gitignore

codeaholics:


mattgodbolt:
Thanks - I had to google this but seems totally sensible to have in the ignore :)

PR #359, opened 2022-06-07
Updated configuration screen

robsonjb1:
Update for the configuration screen to allow more dynamic configuration options.

The user selects from four base models:
	B-DFS1.2 (default)
	B-DFS0.9
	B1770
	Master

Next, the keyboard layout and any optional peripherals are selected:
	6502 co-processor
	Teletext
	Music 5000

The peripherals selected are exposed via URL params.

The default model shown to the user is now BBC DFS 1.2 (rather than 0.9). DFS 1.2 supports the Tube, and I thought it best that the default model would allow any peripheral to be chosen. 

Some 'legacy' models given in the URL have a custom mapping to the new format:

- “MasterTurbo” becomes “Master” + co-pro
- “BTeletext” becomes “B-DFS1.2” + teletext
- “BMusic5000” becomes “B-DFS1.2” + music 5000
- “B” becomes “B-DFS1.2” ?

### DIFF
--- a/config.js
+++ b/config.js
@@ -5,10 +5,14 @@ import { findModel } from "./models.js";
 export function Config(onClose) {
     let changed = {};
     this.model = null;
+    this.coProcessor = null;
     const $configuration = document.getElementById("configuration");
     $configuration.addEventListener("show.bs.modal", () => {
         changed = {};
         setDropdownText(this.model.name);
+        this.set65c02(this.model.tube);
+        this.setTeletext(this.model.hasTeletextAdaptor);
+        this.setMusic5000(this.model.hasMusic5000);
     });
 
     $configuration.addEventListener("hide.bs.modal", () => onClose(changed));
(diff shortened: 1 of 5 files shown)


config.js
```diff
@@ -43,4 +89,40 @@ export function Config(onClose) {
...
+            this.model.os.push(romName);
+        } else {
+            let pos = this.model.os.indexOf(romName);
+            if (pos != -1) {
+                this.model.os.splice(pos, 1);
+            }
+        }
+    };
+
+    this.mapLegacyModels = function (parsedQuery) {
```
mattgodbolt, line 104: Thank you! Perfect!

main.js
```diff
@@ -1127,9 +1136,9 @@ $("#soft-reset").click(function (event) {
 });
 
 function guessModelFromUrl() {
-    if (window.location.hostname.indexOf("bbc") === 0) return "B";
+    if (window.location.hostname.indexOf("bbc") === 0) return "B-DFS1.2";
```
mattgodbolt, line 1139: Love it, thanks! the terrible hack that means `master.godbolt.org` should select the right model :)

main.js
```diff
@@ -1494,7 +1504,7 @@ function stop(debug) {
     }
 
     window.onresize = resizeTv;
-    resizeTv();
+    window.setTimeout(resizeTv, 500);
```
mattgodbolt, line 1507: Is this the hopeful fix for the "my tv is too small?" Maybe we can reduce this number, else there's a half second with the wrong-sized TV, right? Either that or have both a `, 1)` AND a `, 500)`

models.js
```diff
@@ -55,57 +53,27 @@ const masterSwram = [
...
-        beebSwram,
-        I8271,
-        null,
-        true
-    ),
-    new Model(
-        "BBC B (with Music 5000)",
-        ["BMusic5000"],
-        ["os.rom", "BASIC.ROM", "b/DFS-0.9.rom", "ample.rom"],
+        "BBC B with DFS 1.2",
```
mattgodbolt, line 58: much nicer!

mattgodbolt (approved):
Looks great! Thanks so much! I'll merge straight away but if you have followups on the thing I mentioned I'm all ears for more PRs! You're awesome!

mattgodbolt:
This didn't get pushed live as it fails tests; I'm looking at why now.

mattgodbolt:
(easy fix! all good now I think)

dy-tray:
Nice!

PR #361, opened 2022-06-08
update to the 'tiny screen' fix

robsonjb1:
Changes to the `window.setTimeout` logic as per conversation with Matt.

** I was originally using a delay of `, 1` but moved to `, 500` to improve the success from ~95% to 100%. 
So if you agree I have included two bites at the resize at both 1 and 500 intervals as you suggested in order to avoid any initial delay.
(NB: The other change to move the read of the navbar height into the resize function is still needed).

### DIFF
--- a/main.js
+++ b/main.js
@@ -1504,6 +1504,7 @@ function stop(debug) {
     }
 
     window.onresize = resizeTv;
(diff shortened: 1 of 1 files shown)


mattgodbolt (approved):
Thanks!

PR #363, opened 2022-07-02
A collection of (mostly) embed mode improvements

ribbons:
Since automatic screen resizing was implemented, the 'embed' mode in `iframe`s hasn't handled sizing of the screen particularly well.  One of the commits for this PR effectively removes the reserved space around the monitor when in embed mode so that the screen fills as much of the `iframe` as possible.

Related to this, another of the commits sets the audio warning banner to float over the top of the monitor instead of displacing it as this avoids having to re-do the size calculations when the reserved space is removed.  As I personally think this gives a better experience all round I've made it apply outside of 'embed' mode too.

While playing around with the handling of the `iframe` sizing in https://8bs.nerdoftheherd.com/ I also realised it would look tidier if jsbeeb didn't have a different page background colour to the parent page, so another of the commits unsets this when in embed mode.  This also lead to needing to tweak the CUB monitor surround image slightly too.

The other commit just removes an obsolete line of code.

More details in the commit messages to avoid repeating them all here.  Please let me know if I've overlooked something or you're not keen on some of the changes :smile:.


README.md
```diff
@@ -34,7 +34,7 @@ You can then run `npm start` to get a local webserver which will autoreload on c
 - `autorun` - types `*TAPE` then `*/` to run from tape. In conjunction with `loadBasic` it types `RUN`.
 - `autochain` - types `*TAPE` then `CH.""` to run from tape.
 - `autotype` - types whatever you put after. e.g. `&autotype=PRINT"Matt is cool"%0a` (return is URI escaped to `%0a`)
-- `embed` - Adjust the navigation entries to make the page clearer within a 921x733 iframe in a third-party site.
+- `embed` - Remove the margins around the screen, hide most navigation entries and make the page background transparent (intended for use when running within an iframe in a third-party site).
```
mattgodbolt, line 37: Thanks for updating!

mattgodbolt (approved):
Seems like it should work just fine! thanks for this!!

ribbons:
Brilliant, thanks Matt - don't think I've ever had a PR merged as fast :laughing:

mattgodbolt:
You were _very_ lucky :D Normally these things languish a good while...

PR #364, opened 2022-07-03
Make legacy model handling case insensitive too

ribbons:
(Only spotted this as it turned out I was using `model=b` for 8bs.nerdoftheherd.com)

Handling of the current model names is case insensitive, e.g. `model=master` is valid as well as `model=Master`.  `mapLegacyModels` is however case sensitive so `model=B` works but `model=b` causes an uncaught `TypeError`.

Resolve this inconsistency by making `mapLegacyModels` case insensitive.  Also took the opportunity to move handling of the legacy `B` model to just be another alias of `BBC B with DFS 0.9` and corrected a comment.

### DIFF
--- a/config.js
+++ b/config.js
@@ -102,27 +102,26 @@ export function Config(onClose) {
     };
 
     this.mapLegacyModels = function (parsedQuery) {
+        if (!parsedQuery.model) {
+            return;
+        }
+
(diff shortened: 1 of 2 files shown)


mattgodbolt (approved):
thanks!

PR #367, opened 2022-08-04
Pass disc reference to flusher().

regregex:
Fixes use of browser local storage for discs in Firefox 102.0. Writing to disc failed because `this` was undefined in the function referenced by `BaseDisc.flusher()`, and the `BaseDisc.data` member could not be retrieved.

Examples of failing URLs:
[*FORM 80 0](https://bbc.godbolt.org/?model=Master&disc1=local:foo&autotype=%2AFORM%2080%200%0AYYY)
[SAVE "BAR"](https://bbc.godbolt.org/?model=Master&disc1=local:bar&autotype=SAVE%22BAR%22%0A)

The menu option to download the disc image is enabled in all cases, to mitigate potential dataloss if local storage is cleared unexpectedly (in my configuration it is happening on browser close, for localhost only.)

### DIFF
--- a/fdc.js
+++ b/fdc.js
@@ -44,10 +44,10 @@ export function discFor(fdc, name, stringData, onChange) {
         return res;
     }
 
-    return new BaseDisc(fdc, name, data, () => {
+    return new BaseDisc(fdc, name, data, (disc) => {
         if (!changed()) return;
         if (onChange) {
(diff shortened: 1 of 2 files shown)


mattgodbolt (approved):
LGTM! thanks for fixing this!

PR #368, opened 2022-08-04
Set 3:1 Teletext flash ratio.

regregex:
Adjusts the flash periods of flashing text and graphics in `MODE 7` to 48 fields on, 16 fields off, and has them extinguished at the same time as a slow-flashing cursor. Matching the behaviour of the SAA 5050 and MC6845.

Synchrony is not critical: as the two chips count fields independently and in lockstep, a `MODE` change typically preserves their respective counts but may occasionally desynchronise them by one field.


mattgodbolt (approved):
I'll take your word on this! Thanks so much :)

PR #392, opened 2023-03-28
Update cmos.js

robsonjb1:
Catch an invalid FS configuration setting (possibly as a result of using a prior version that didn't set a default of 254 in CMOS)


mattgodbolt (approved):
Thanks :)

PR #394, opened 2023-07-01
Configure tests, extract query parsing

mattgodbolt:
Limited start to testing some of the monolith of the main js file.


mattgodbolt:
Obsoleted by #483

PR #400, opened 2023-08-17
Updated Tube ULA implementation

samskivington-poq:
Hello.

I've updated the Tube ULA implementation again.

It's now much more readable, making it easier to understand, and (hopefully never have to, but...) work out what's going on if it turns out I've missed something.

Added the functionality for two bits when writing to the R1 status register.

Re-added the incorrectly removed check for parasite to host R3 being empty when deciding whether a parasite NMI should be asserted.

Adjusted a few bits of logic as well to match observed behaviour on real hardware, and information in the Tube Application Note.

Please let me know your thoughts.

:)

### DIFF
--- a/tube.js
+++ b/tube.js
@@ -128,9 +128,6 @@ export class Tube {
         this.parasiteCpu.resetHeldLow = (this.internalStatusRegister & TUBE_ULA_FLAG_STATUS_PARASITE_RESET_ACTIVE_LOW);
     }
     hostRead(address) {
-        //  Not implemented - needs to be integrated with the parasite CPU code:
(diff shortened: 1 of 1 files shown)


mattgodbolt:
I'm so sorry this fell between the cracks!

tube.js
```diff
@@ -1,231 +1,353 @@
 "use strict";
 import * as utils from "./utils.js";
 
+//  this one should be declared more globally
+const HOST_CPU_FLAG_IRQ_TUBE_ULA = 8;
+
+const TUBE_ULA_R1 = 0;
```
mattgodbolt, line 7: I love the extraction of these constants to named values. thanks!

PR #427, opened 2024-08-10
Changes to use typed arrays for 8/16 bit stuff

mattgodbolt:
Not sure this is better.


mattgodbolt:
You're honestly absolutely on the money there Claude :)

I couldn't measure any performance difference and it makes the code more ugly.

Issue #461, opened 2024-12-22
Tests fail with vite


codeaholics:
With commit `4a83d43`, tests fail.

```
% make test

[...]

 Test Files  2 failed | 7 passed (9)
      Tests  7 failed | 35 passed | 2 skipped (44)
   Start at  21:13:09
   Duration  6.41s (transform 222ms, setup 0ms, collect 1.88s, tests 17.93s, environment 1ms, prepare 533ms)
```

All `ENOENT` failures, for example: `Error: ENOENT: no such file or directory, open 'tests/integration/dp111_6502Timing/65C12timing.ssd'`

codeaholics:
Ignore me. I forgot to fetch the submodules. Is there a way to make this easier? Not using submodules, or explicitly checking if they're initialised before running the tests? Or even a test that fails with an explicit error message reminding you to do the submodules dance if the files are missing?

codeaholics:
I opened a PR with a test to ensure the submodules are present and a useful error message if not.

mattgodbolt:
Thanks for chasing this one down!! :)

PR #462, opened 2024-12-22
Add tests to ensure submodules are present and fail with a more explicit error message if not

codeaholics:
See #461

### DIFF
--- a/tests/integration/ensure-submodules.js
+++ b/tests/integration/ensure-submodules.js
@@ -1,0 +1,21 @@
+import { describe, it } from "vitest";
+import assert from "assert";
+import * as fs from "fs";
+
+describe("ensure git submodules are present", function () {
+    it("should have functional tests", function () {
+        try {
+            fs.accessSync("tests/6502_65C02_functional_tests/README.md");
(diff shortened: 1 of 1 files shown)


tests/integration/ensure-submodules.js
```diff
@@ -0,0 +1,21 @@
+import { describe, it } from "vitest";
+import assert from "assert";
+import * as fs from "fs";
+
+describe("ensure git submodules are present", function () {
+    it("should have functional tests", function () {
+        try {
+            fs.accessSync("tests/6502_65C02_functional_tests/README.md");
+        } catch {
+            assert.fail("Functional tests submodule missing. Ensure git submodules are fetched.");
```
mattgodbolt, line 10: fantastic! thanks for this: perhaps add "Try something like `git submdule update --init`" for those who don't remember the incantation.

mattgodbolt (approved):

tests/integration/ensure-submodules.js
mattgodbolt: I'll update after merging

PR #471, opened 2025-03-11
Fixed the Dockerfile to work with Node 20.

pacman-ghost:


package.json
```diff
@@ -8,6 +8,7 @@
   "description": "Emulate a BBC Micro",
   "repository": "git@github.com:mattgodbolt/jsbeeb.git",
   "version": "0.0.7",
+  "//": "If you change the version of Node, it must also be updated at the top of the Dockerfile.",
```
mattgodbolt, line 11: Is this ok? I'll merge, and hope it is...

pacman-ghost: This is [apparently](https://groups.google.com/g/nodejs/c/NmL7jdeuw0M/m/yTqI05DRQrIJ) how you comment a lock file.
```
That's why the "//" key will never be used by npm for any purpose, and is reserved for comments.
```

How someone can, in this day and age, come up with a file format that doesn't allow for comments is a rant for another day... :-/

mattgodbolt: 100% with you there! TIL you can "comment" but sadly not in a way that is useful. mostly I want to comment package.json with:
```js
{  "bob": "^2.3.4", // Don't upgrade this, newer versions have a bug we rely on ...
...
```

But hey ho :) Thanks!

PR #522, opened 2025-10-16
I let OpenAI Codex try and fix #326!

### LINKED ISSUE #326: Support "proper" address translation
https://github.com/mattgodbolt/jsbeeb/blob/a2cf2763a266a3b19b0d43d97231b569fa18527b/video.js#L245-L264

c.f. http://beebwiki.mdfs.net/Address_translation

codeaholics:
I'm experimenting with OpenAI Codex. I asked it to fix #326 with very limited context. It read the referenced beebwiki page and came up with this fix. I'm slightly suspicious that the 4 and 10 on line 156 might need to be swapped, but I'm not sure. I also don't know of any BBC games that fail because of this issue.

Take it or leave it, I won't be offended! I'm having fun! :) 

(PS: Elite and Exile appear to work.)

### DIFF
--- a/src/via.js
+++ b/src/via.js
@@ -620,7 +620,7 @@ export class SysVia extends Via {
         this.capsLockLight = !(this.IC32 & 0x40);
         this.shiftLockLight = !(this.IC32 & 0x80);
 
-        this.video.setScreenAdd((this.IC32 & 16 ? 2 : 0) | (this.IC32 & 32 ? 1 : 0));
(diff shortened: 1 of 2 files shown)


codeaholics:
Trueeeeee... but... they worked before! I found some games that are alleged to have hardware scrolling - most notably Repton. And it worked before (and after). So it's really hard to actually say this has fixed anything. I might have another go at it, this time without my automated helper. Something that at the very least makes reference to real hex screen addresses, or sizes, or offsets would at least be more readable. Maybe even some tests.

PR #526, opened 2025-11-07
Fix precision problems in PAL mode

mattgodbolt:
- `precision highp float;` - we need more than 10 bits for the
   maths we're doing. My laptop defaulted to "23 bits" for medium
   precision, so that hid the issues until I tried it on latest
   NVidia drivers on my desktop.
- Using gl_FragCoord instead of a varying. Probably unnecessary
  but we can avoid a varying this way. Claude reckons:
  "Avoids barycentric interpolation artifacts" but that might be
  FUD honestly, we experimented a bunch to work out what was going
  on and this didn't seem like a _bad_ change, so we left it in.

### DIFF
--- a/src/video-filters/shaders/pal-composite.frag.glsl
+++ b/src/video-filters/shaders/pal-composite.frag.glsl
@@ -1,9 +1,9 @@
-precision mediump float;
+precision highp float;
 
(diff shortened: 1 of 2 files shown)


mattgodbolt:
@claude take a look please

PR #540, opened 2025-11-25
feat: Optimised polltime routine

mattgodbolt:
Seemed to give a consistent bump in performance (albeit on battery power on a plane)


mattgodbolt:
@claude what do you reckon?

PR #549, opened 2026-01-24
fix: decrement SP by 3 during reset sequence

### LINKED ISSUE #547: SP not decremented during reset sequence (should be 0xFD, not 0x00)
### Description

After reset, the 6502 stack pointer should be 0xFD, but jsbeeb leaves it at its initial value (0x00).

### Background

The real 6502 reset sequence is 7 cycles and reuses the interrupt/BRK logic internally. During this sequence, 3 "dummy" stack operations occur where SP decrements but R/W is held in read mode (so no actual writes happen). This was a hardware behaviour to avoid corrupting memory at power-on when SP contains random garbage.

Result: SP decrements 3 times during reset: 0x00 → 0xFF → 0xFE → 0xFD

### Reproduction

```javascript
import { describe, it, expect } from "vitest";
  import { fake6502, fake65C12 } from "../../src/fake6502.js";

  describe("6502 reset SP behavior", () => {
      it("should decrement SP by 3 during reset (NMOS)", async () => {
          const cpu = fake6502();
          await cpu.initialise();
          // Real 6502 does 3 dummy pushes: 0x00 → 0xFF → 0xFE → 0xFD
          expect(cpu.s).toBe(0xfd);
      });

      it("should decrement SP by 3 during reset (CMOS)", async () => {
          const cpu = fake65C12();
          await cpu.initialise();
          expect(cpu.s).toBe(0xfd);
      });

      it("should have SP=0xFD after explicit reset call", async () => {
          const cpu = fake6502();
          await cpu.initialise();
          cpu.s = 0x80;
          // Reset should decrement by 3: 0x80 → 0x7F → 0x7E → 0x7D
          cpu.reset(true);
          expect(cpu.s).toBe(0x7d);
      });
  });
```

Save as `tests/unit/test-sp-reset.js` and run with `npx vitest run tests/unit/test-sp-reset.js`

### Impact

Stack operations (JSR, PHA, interrupts) write to addresses offset by 3 bytes from real hardware:
  - With SP=0x00: First push goes to $0100, SP wraps to 0xFF
  - With correct SP=0xFD: First push goes to $01FD, SP becomes 0xFC

This causes the stack page ($0100-$01FF) to have data at different offsets than real hardware. If code run shortly after reset (*e.g.* `LDX #$FF; TXS`) initialises the SP deterministically, you may never notice any difference in practice. On the other hand, if you don't ... 

### Suggested Fix

In `src/6502.js reset()` method, after setting PC, add:

```javascript
this.s = (this.s - 3) & 0xff;  // Simulate 3 dummy pushes
```

### References

  - https://www.nesdev.org/wiki/CPU_power_up_state — NES CPU power-up state documentation: "S -= 3" during reset, resulting in $FD
  - https://www.pagetable.com/?p=410 — "Internals of BRK/IRQ/NMI/RESET on a MOS 6502"
  - https://superuser.com/a/606770
  - http://forum.6502.org/viewtopic.php?p=2959 — "Stack Pointer and Register Status After RESET"

rob-smallshire:
## Summary

Fixes #547

The real 6502 reset sequence reuses the interrupt/BRK logic internally, performing 3 "dummy" stack operations where SP decrements but R/W is held in read mode (no actual writes). This results in SP being 0xFD after reset when starting from 0x00.

This PR adds the missing SP decrement to both `Cpu6502.reset()` and `Tube6502.reset()` methods.

## Changes

- `src/6502.js`: Added `this.s = (this.s - 3) & 0xff` after setting PC and I flag in both reset methods
- `tests/unit/test-sp-reset.js`: New test file covering SP reset behavior

## Test plan

- [x] New unit tests pass (5 tests covering NMOS/CMOS, hard/soft reset, wrap-around)
- [x] Full unit test suite passes (255 tests)
- [x] Integration tests pass (45 tests)
- [x] CPU compatibility test suite passes
- [x] Linting passes

### DIFF
--- a/src/6502.js
+++ b/src/6502.js
@@ -419,6 +419,7 @@ class Tube6502 extends Base6502 {
         this.romPaged = true;
         this.pc = this.readmem(0xfffc) | (this.readmem(0xfffd) << 8);
         this.p.i = true;
+        this.s = (this.s - 3) & 0xff; // Simulate 3 dummy pushes during reset
         this.tube.reset(hard);
     }
 
@@ -1181,6 +1182,7 @@ export class Cpu6502 extends Base6502 {
         }
         this.pc = this.readmem(0xfffc) | (this.readmem(0xfffd) << 8);
         this.p.i = true;
+        this.s = (this.s - 3) & 0xff; // Simulate 3 dummy pushes during reset
         this._nmiEdge = false;
         this._nmiLevel = false;
         this.halted = false;
--- a/tests/unit/test-sp-reset.js
+++ b/tests/unit/test-sp-reset.js
@@ -1,0 +1,59 @@
+import { describe, it, expect } from "vitest";
+import { fake6502, fake65C12 } from "../../src/fake6502.js";
+
+/**
+ * Tests for 6502 stack pointer behavior during reset.
+ *
+ * The real 6502 reset sequence is 7 cycles and reuses the interrupt/BRK
+ * logic internally. During this sequence, 3 "dummy" stack operations occur
+ * where SP decrements but R/W is held in read mode (no actual writes).
+ * This was a hardware optimization to share logic with BRK/IRQ handling.
+ *
+ * Result: SP decrements 3 times during reset: 0x00 -> 0xFF -> 0xFE -> 0xFD
+ *
+ * References:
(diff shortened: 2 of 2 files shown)


mattgodbolt (approved):
Love it! thanks so much, and for the tests too 🎉

PR #565, opened 2026-02-26
feat: wire up accessibility switch keys for user port, ADC, and fire buttons

### LINKED ISSUE #160: USER Port access (especially for switch interface box and joystick Special Needs access)
Adding USER PORT emulation would facilitate access to a lot of BBC Micro software designed for disabled people unable to use conventional controls. Typically this was via a switch interface box with four sockets, or a joystick. This link explains a little more from a Brilliant Computing catalogue.

[https://www.dropbox.com/s/idmzhdv84rmyr5v/09-10.jpg?dl=0](url)

I believe it's an 8-line interface, which writes the status to memory location ?&FE60 as follows....
​
?&FE60 = 255 11111111 = no switches pressed via the USER PORT
?&FE60 = 254 11111110 = Switch 1 / Right pressed via the USER PORT
?&FE60 = 253 11111101 = Switch 2 / Left pressed via the USER PORT 
?&FE60 = 251 11111011 = Switch 3 / Up pressed via the USER PORT
?&FE60 = 247 11110111 = Switch 4 / Down pressed via the USER PORT

.... and for more than one at a time, working as this example: 252 11111100 = Switch 1 + 2 together..... Quite likely that some activities may use a fifth switch for a fire button. 

Here's some software you can test this with....

Thurrock Care Switch programmes: If you select the talking word processor, you can test all 4 switches quite easily. https://bbc.godbolt.org/?autoboot&disc=https://dl.dropboxusercontent.com/s/8pralcrfztknabu/DISC000.ssd?dl=0

Joystick Games: Select USER PORT for connection. As you can see, these lock up because there's no way to reset the analogue or user ports. They both seem to have values jamming them on. https://bbc.godbolt.org/?autoboot&disc=https://dl.dropboxusercontent.com/s/zetb1jcf9julkmq/DISC013.ssd?dl=0

Spot on Games: on-line. Also locks up before you can start it... https://bbc.godbolt.org/?autoboot&disc=https://dl.dropboxusercontent.com/s/rz1s4i2j3vtnvhx/DISC017.ssd?dl=0

It may be necessary to set the user-port to 255 (no switches pressed) and also to "reset" the analogue port to get stuff all working, as some titles will lock up before starting if they think you are holding a control down.

Fingers crossed for one day.

mattgodbolt-molty:
Fixes #160.

## What was wrong

The existing user port key mapping (added 2018, commit c77e85b) was dead code. It set up handlers in a local `emuKeyHandlers` map that was never connected to the keyboard system after the 2025 keyboard refactor (commit 64786cf), so pressing any key never actually updated `&FE60`.

Additionally, `keyboard.js` called `sysvia.keyDown()` before checking registered handlers, so any handled key also leaked through to the emulated machine.

## Hardware archaeology

From the Brilliant Computing catalogue (the original supplier referenced in the issue):

- **Interface Box** — connects to the User Port via ribbon cable; accepts up to 4 switches with ¼" jack plugs
- **Special Ed. Joystick** — also User Port; description explicitly says *"Plugs straight into the 'User Port' (no Interface Box needed)"*

Both devices affect **only** `?&FE60`. The analogue port (ADC) and System VIA fire buttons (PB4/PB5) belong to the *standard* analogue joystick connector — a completely separate physical port. Real switch hardware never touches them.

An earlier version of this PR also deflected ADC channels and triggered fire buttons alongside user port bits, following a commenter's ideal spec. Testing with Thurrock Care showed this was wrong — coupling Alt+1 to PB4 caused an unwanted "select current item" action to fire every time a switch was pressed.

### `src/keyboard.js`
- Check registered handlers *first* in `keyDown()`; if one fires, return early without calling `sysvia.keyDown()`. Handled keys no longer leak through to the BBC.
- `keyUp()` still always forwards to `sysvia.keyUp()` to avoid sticky keys.

### `src/main.js`
- Remove the orphaned `emuKeyHandlers` map and dead `switchKey` loop
- Register switch handlers via `keyboard.registerKeyHandler()` after keyboard initialisation
- **Alt+1–8** → user port bits 0–7 (active low). Restores the original 2018 intent; normal number keys reach the BBC unaffected.
- **Alt+F1–F8** → same switch bits, ergonomic alternative; normal BBC function keys (no Alt) are completely unaffected.
- Nothing else — no ADC deflection, no fire button coupling. Matches real hardware.

### `tests/unit/test-keyboard.js`
- Add assertion to existing Alt-handler test that `sysvia.keyDown` was *not* called
- Two new tests: handler suppresses `sysvia.keyDown`; unhandled keys still reach it

## Switch mapping

| Key | User port bit | `?&FE60` when pressed |
|-----|--------------|----------------------|
| Alt+F1 / Alt+1 | 0 | 254 |
| Alt+F2 / Alt+2 | 1 | 253 |
| Alt+F3 / Alt+3 | 2 | 251 |
| Alt+F4 / Alt+4 | 3 | 247 |
| Alt+F5 / Alt+5 | 4 | 239 |
| Alt+F6 / Alt+6 | 5 | 223 |
| Alt+F7 / Alt+7 | 6 | 191 |
| Alt+F8 / Alt+8 | 7 | 127 |

## Testing

257 unit tests pass including 2 new keyboard tests. Manual testing with the test discs from the issue:
- [Thurrock Care switches](https://bbc.godbolt.org/?autoboot&disc=https://dl.dropboxusercontent.com/s/8pralcrfztknabu/DISC000.ssd?dl=0) — talking word processor, Alt+1–4 should navigate without triggering "select"
- [Joystick Games](https://bbc.godbolt.org/?autoboot&disc=https://dl.dropboxusercontent.com/s/zetb1jcf9julkmq/DISC013.ssd?dl=0) — select USER PORT mode, Alt+1–4 for directions

*(I'm Molty, an AI assistant acting on behalf of @mattgodbolt)*


oneswitch:
Ah magic. I'll bury into this properly ASAP, but I can see already it's going to bring those things back to life. Thanks so much!

PR #597, opened 2026-03-22
fix: correct 65C12 timing for dead cycles, RMW spurious ops, and ACCCON TST

### LINKED ISSUE #596: Incorrect timings 65c02 vs 6502
@tom-seddon has some test cases https://github.com/tom-seddon/beeb_6502_timing_tests that pass on NMOS 6502s but fails a few of the 65c02 tests. More info: https://github.com/bitshifters/bbc-documents/blob/master/ICs/6502%20CPU/6502%20Dead%20Cycles.pdf

We should incorporate Tom's tests into our test system and fix them for the 65c02.

6502:

Image

65c02:

Image

mattgodbolt-molty:
## Summary

Fixes three 65C12 (BBC Master) CPU timing issues exposed by [Tom Seddon's timing tests](https://github.com/tom-seddon/beeb_6502_timing_tests):

- **`STA (),Y` dead cycle**: On the 65C12, the dead cycle in indirect indexed stores accesses the Previous Bus Address (always 2MHz), not the target address. Changed from `readOp(addrNonCarry)` to `tick(1)` to avoid spurious 1MHz bus stretching.

- **`abs,X` RMW spurious cycle**: The 65C12 does "two reads and one write" for RMW instructions, not the 6502's "one read and two writes". Changed the explicit `writeOp` to `readOp` for the spurious cycle, matching how `abs` mode already handles this via `spuriousOp`.

- **ACCCON TST bit**: When the Master's ACCCON bit 6 is set, FRED/JIM (&FC00-&FDFF) **reads** are remapped to internal 2MHz bus, but **writes** still go through the external 1MHz bus. `polltimeAddr` now distinguishes reads from writes so TST only exempts reads from 1MHz stretching.

Also adds Tom Seddon's test disc as an integration test covering dead cycles, page boundary crossing, and 1MHz bus timing edge cases for both 6502 and 65C12.

Fixes #596

## Test plan

- [x] Tom Seddon's timing tests pass for both BBC B (6502) and Master (65C12)
- [x] All 401 unit tests pass
- [x] All 50 integration tests pass (including dp111 timing, RMW, Dormann functional)
- [x] CPU compatibility tests pass
- [x] Lint passes

### DIFF
--- a/src/6502.js
+++ b/src/6502.js
@@ -1307,13 +1307,14 @@ export class Cpu6502 extends Base6502 {
     }
 
     // Bake in the 1MHz bus check at construction time. On the Master, ACCCON
-    // TST (bit 6) remaps FRED/JIM reads to the internal 2MHz bus, but writes
-    // still go through the external 1MHz bus. Non-Master machines just use the
-    // static address check with no ACCCON awareness.
+    // TST (bit 6) remaps FRED, JIM, and SHEILA reads (&FC00-&FEFF) to the
+    // internal 2MHz bus, but writes still go through the external 1MHz bus.
+    // Non-Master machines just use the static address check with no ACCCON
+    // awareness.
     buildIs1MHzAccess() {
         if (this.model.isMaster) {
             return (addr, isWrite) => {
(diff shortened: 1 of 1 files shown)


mattgodbolt:
CC @tom-seddon

PR #640, opened 2026-04-04
refactor: standardise async patterns in main.js

mattgodbolt:
## Summary

Replace all `.then()` patterns in main.js with consistent async/await:

- Remove no-op `.then(() => {})` on fire-and-forget promises (setupMicrophone, tryResume, ensureMicrophoneRunning)
- Convert `.then(fn)` chains to async/await in Google Drive and disc list click handlers
- Convert top-level `startPromise.then().catch()` to async IIFE with try/await/catch

Zero `.then()` calls remain in main.js.

## Test plan

- [x] All 485 tests pass
- [x] Production build succeeds
- [ ] Disc loading from built-in list works
- [ ] Google Drive disc loading works

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -1382,8 +1382,7 @@ for (const image of availableImages) {
         utils.noteEvent("images", "click", image.file);
         setDisc1Image(image.file);
         $discsModal.hide();
-        const disc = await loadDiscImage(parsedQuery.disc1);
-        processor.fdc.loadDisc(0, disc);
+        processor.fdc.loadDisc(0, await loadDiscImage(parsedQuery.disc1));
     });
 }
 


src/main.js
```diff
@@ -1361,13 +1361,12 @@ googleDriveEl.addEventListener("show.bs.modal", async function () {
...
-        row.addEventListener("click", function () {
+        row.addEventListener("click", async function () {
             utils.noteEvent("google-drive", "click", item.name);
             setDisc1Image(`gd:${item.id}/${item.name}`);
-            gdLoad(item).then(function (ssd) {
-                processor.fdc.loadDisc(0, ssd);
-            });
             googleDriveModal.hide();
+            const ssd = await gdLoad(item);
+            if (ssd) processor.fdc.loadDisc(0, ssd);
```
mattgodbolt, line 1369: can ssd be falsey here? we didn't previously handle this

src/main.js
```diff
@@ -1379,13 +1378,12 @@ for (const image of availableImages) {
...
     elem.querySelector(".description").textContent = image.desc;
-    elem.addEventListener("click", function () {
+    elem.addEventListener("click", async function () {
         utils.noteEvent("images", "click", image.file);
         setDisc1Image(image.file);
-        loadDiscImage(parsedQuery.disc1).then(function (disc) {
-            processor.fdc.loadDisc(0, disc);
-        });
         $discsModal.hide();
+        const disc = await loadDiscImage(parsedQuery.disc1);
```
mattgodbolt: ditto falsiness

src/main.js
mattgodbolt: Yes — gdLoad catches errors internally (shows error dialog, calls loadingFinished) and returns undefined in that case. The falsy check preserves that.

src/main.js
mattgodbolt: loadDiscImage always returns a disc or throws, so this cannot be falsy. No guard needed — matches the original behaviour.

src/main.js
mattgodbolt: Actually, correction: the original .then() code would have called loadDisc(0, undefined) on error — the falsy check is a bug fix, not just preservation of behaviour.

Issue #674, opened 2026-04-08
Atom tests


mattgodbolt:
We should do manual testing of the following: (via @commandercoder)

For software the MMC I used was created by Hoglet from Stardot.
https://github.com/hoglet67/AtomSoftwareArchive/releases

https://github.com/hoglet67/AtomSoftwareArchive/releases/download/V14BETA2/AtomSoftwareArchive_20260227_0939_V14BETA2.zip

I used the games here for testing graphics modes:

```
            0xf0: [8, 1, 1, 1], //clear4  256x192x2,  pixels 1w1h   MAIN MENU
            0xb0: [16, 2, 1, 1], //clear3  128x192x2, pixels  2w1h   BABIES
            0x70: [16, 2, 2, 1], //clear2  128x96x2, pixels  2w2h  3D ASTEROIDS
            0x30: [16, 2, 3, 1], //clear1   128x64x2 , pixels  3w4h (2w4h) 3D MAZE
            0xd0: [8, 2, 1, 2], //?#B000=#D0  128x192x4,pixels  2w1h CHUCKIE EGG
            0x90: [8, 2, 2, 2], //?#B000=#90  128x96x4, pixels 2w2h  FLAPPY BIRD
            0x50: [8, 2, 3, 2], //?#B000=#50  128x64x4 , pixels 3w3h (4w3h) BREAKOUT (maingame)
            0x10: [16, 4, 3, 2], //?#B000=#10  64x64x4 , pixels 4w3h  FIZZLE BRICKS
            0x00: [8, -1, 12, 1], // clear0 //0,0 not used on Mode 0 (uses blitChar), pixelsPerBit, bpp
```

I used GAGS to test the bank switching.

My version of *jsatom* is on my site for comparison.  Press CTRL-BREAK to
get the menu.
https://atom.commandercoder.com/

PR #680, opened 2026-04-12
fix: Atom tests, PPIA mirroring, and speaker timing

### LINKED ISSUE #658: fix: PPIA 8255 address mirroring (incomplete decode)
The 8255 PPI has incomplete address decoding, so registers 0-3 mirror at addresses 4-7. Currently:
- Reads from 0xB004-0xB007 return open bus instead of mirroring 0xB000-0xB003
- Writes to 0xB004-0xB007 pass through to \`PPIA.write(addr)\` which uses \`addr & 0xf\`, giving register numbers 4-7 that hit the default/CREG case

The fix would be to mask addresses to \`addr & 0x3\` before passing to the PPIA, or have the PPIA internally mirror registers 4-7 to 0-3.

This matches the original PR #505 behaviour so it's not a regression, but it's technically incorrect per the 8255 datasheet.

Low priority -- unlikely to affect any real Atom software.

### LINKED ISSUE #670: fix: speaker timing uses sample indices instead of CPU cycle timestamps
## Problem

In `AtomSoundChip.speakerChannel()`, the transition timing comparison mixes units:

- `this.bitChange[].cycles` stores **CPU cycle timestamps** (from `updateSpeaker`: `microCycle + seconds / this.secondsPerCycle`)
- `fromTime` is computed as `this.scheduler.epoch - length`, where `length` is a **sample count**
- The loop compares `this.bitChange[i].cycles <= fromTime + i` where `i` is a **sample index**

Since `samplesPerCycle` is 0.5 for the Atom at 1 MHz (500 kHz sample rate / 1 MHz CPU), this effectively treats 1 sample as 1 cycle, which is off by a factor of 2. The timing works empirically because the original implementation was tuned to compensate, but it's not technically correct.

## Expected behaviour

Transition timestamps should be compared in the same units. Either:
- Convert sample indices to cycle timestamps: `fromCycle + i / samplesPerCycle`
- Or store transitions in sample-time units in `updateSpeaker`

## Impact

Low. The speaker produces audible sound and the timing is "close enough" for the square waves the Atom generates. A fix would improve timing accuracy for programs that rely on precise speaker toggle rates for specific frequencies.

## Context

This was inherited from CommanderCoder's original implementation in PR #505 and has been flagged by Copilot in multiple reviews (#647, #669). Deferred from those PRs to avoid breaking working audio.

mattgodbolt-molty:
## Summary

Three Atom issues, all TDD'd with failing tests written first:

- **PPIA address mirroring (#658)**: The 8255 only decodes A0-A1, so registers 0-3 mirror at 4-7. Changed `addr & 0xf` to `addr & 0x3`, with a guard to ignore addresses 0xB008-0xB00F (outside the 8255's decode range).
- **Speaker timing (#670)**: `speakerChannel()` subtracted a sample count from a CPU-cycle epoch and compared cycle timestamps against sample indices. With `samplesPerCycle = 0.5`, transitions appeared at the wrong position (off by 2x). Fixed by converting consistently to cycle-space.
- **Atom integration tests (#674)**: Video mode switching, memory layout, PPIA mirroring (via BASIC), and BASIC execution (FOR loops, variable assignment, arithmetic) — all using the headless Atom emulator with VSync simulation.

Fixes #658, fixes #670, partially addresses #674

## Test plan

- [x] `npm run test:unit` — 573 tests pass
- [x] `npm run test:integration` — 65 tests pass (12 new Atom tests)
- [x] PPIA mirroring: 6 unit tests (mirrored addresses + 0xB008 guard)
- [x] Speaker timing: 1 new test verifying transition appears at correct sample index
- [x] Atom hardware: 7 integration tests (mode switching, memory, BASIC)

### DIFF
--- a/src/ppia.js
+++ b/src/ppia.js
@@ -99,7 +99,7 @@ class PPIA {
  */
     write(addr, val) {
         val |= 0;
-        switch (addr & 0xf) {
+        switch (addr & 0x3) {
             case PORTA:
                 this.latcha = val;
                 this.recalculatePortAPins();
@@ -142,7 +142,7 @@ class PPIA {
     }
 
     read(addr) {
-        switch (addr & 0xf) {
+        switch (addr & 0x3) {
             case PORTA:
                 this.recalculatePortAPins();
                 return this.portapins;
--- a/src/ppia.js
+++ b/src/ppia.js
@@ -99,6 +99,9 @@ class PPIA {
  */
     write(addr, val) {
         val |= 0;
+        // The 8255 only decodes A0-A1; addresses 4-7 mirror 0-3.
+        // Addresses 8-15 are outside the 8255's decode range.
+        if ((addr & 0xf) >= 0x8) return;
         switch (addr & 0x3) {
             case PORTA:
                 this.latcha = val;
@@ -142,6 +145,7 @@ class PPIA {
     }
 
     read(addr) {
+        if ((addr & 0xf) >= 0x8) return addr >>> 8; // open bus
         switch (addr & 0x3) {
             case PORTA:
                 this.recalculatePortAPins();
--- a/src/soundchip.js
+++ b/src/soundchip.js
@@ -378,7 +378,7 @@ export class AtomSoundChip extends SoundChip {
     }
 
     speakerChannel(channel, out, offset, length) {
-        let fromTime = this.scheduler.epoch - length;
+        const fromCycle = this.scheduler.epoch - length / this.samplesPerCycle;
         let bitIndex = 0;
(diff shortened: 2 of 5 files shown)


mattgodbolt-molty:
Re: Copilot's speakerChannel chunked-advance concern — this is a pre-existing issue. The old code had the same structural problem (`epoch - length` varies per chunk while epoch stays fixed). The buffer is 4096 samples so chunking rarely triggers for the Atom's short `advance()` calls. Not worth fixing in this PR but could be a follow-up if it causes audible glitches.

PR #689, opened 2026-05-03
feat: load STH discs and tapes from the bbc.xania.org mirror

mattgodbolt-molty:
## Summary

Switches `src/sth.js` from scraping the catalog and zips off
`www.stairwaytohell.com` to the mirror at `bbc.xania.org/archive/sth/`
populated by #688.

* Drops the `DOMParser` HTML scrape of `reclist.php` in favour of a single
  `manifest.json` GET per category. The disc catalogue goes from ~350 KB of
  scraped HTML to ~23 KB on the wire (125,927 bytes of JSON, gzipped by
  CloudFront), so opening the archive browser is roughly 15x lighter.
* URL-encodes each path component, so filenames containing characters like
  brackets (e.g. `Unreleased/Daxis[droids]-demo.zip`) produce a well-formed
  request. The old direct-to-STH flow relied on the upstream server tolerating
  a literal `[` and `]` in the path.
* Always fetches over https rather than following the page's protocol. The
  mirror redirects plain http, so inheriting it cost a 301 on every catalogue
  and disc fetch when running the dev server over http. This also subsumes the
  old `file:` special case for Electron, and `src/sth.js` no longer touches
  `document` at all.
* No fallback to upstream STH — by design, since the mirror is now
  authoritative and the point of this work is to stop depending on the live
  site.

The `sth:` URL syntax (`?disc1=sth:Acornsoft/Elite.zip`) and the in-app modal
are unchanged; manifest paths are spelled exactly as upstream's hrefs spelled
them, so existing saved links keep working. The credit link in the modal still
points at the original stairwaytohell.com site.

## Dependency

Merge after #688. The mirror itself is **already live and populated**, so the
only remaining dependency is the code in that PR landing first:

```sh
curl --compressed https://bbc.xania.org/archive/sth/diskimages/manifest.json | head -c 200
curl --compressed https://bbc.xania.org/archive/sth/tapeimages/manifest.json | head -c 200
```

Use `--compressed`, not `curl -I`: CloudFront doesn't compress HEAD responses,
so a HEAD request reports the origin's uncompressed length and makes it look
like compression is off.

## Test plan

- [x] `npm run ci-checks` clean
- [x] `npx vitest run tests/unit` green, including the 5 tests here covering
      catalog parse, tape vs disc selection, error reporting, manifest-shape
      validation, and `[` / `]` path encoding
- [x] Ran this branch's `src/sth.js` unmodified against the live bucket, with
      no mocked `fetch`:
  - disc `populate()` -> 1,608 entries; tape `populate()` -> 1,502 entries
  - a second `populate()` uses the cached catalog rather than refetching
  - `fetch()` end to end through `unzipDiscImage`: `Elite-MasterAndTubeEnhanced.zip`
    -> a 179,456-byte `.ssd`, and `Elite-v1.0_B.zip` -> a 27,174-byte `.uef`
  - `fetch("Unreleased/Daxis[droids]-demo.zip")` -> a 69,888-byte `.ssd`,
    requested as `Daxis%5Bdroids%5D-demo.zip`
  - a nonexistent path rejects rather than resolving empty
- [x] Manifests serve with `access-control-allow-origin: *`, so this works from
      any origin, not just bbc.xania.org
- [x] Confirmed in a browser against the live mirror (thanks @mattgodbolt): the
      301-per-request behaviour above was spotted this way and is now fixed
- [ ] Re-check the modals in a browser after the https change: open the disc modal,
      filter, click an entry; same for tapes; and load
      `?disc1=sth:Acornsoft/Elite.zip`. The code path underneath is covered
      above, but the modal wiring itself hasn't been exercised in a browser.


mattgodbolt:
The mirror is now live, so the dependency on #688 is satisfied on the infrastructure side — this just needs #688 merged.

Verified against the real bucket:

```
https://bbc.xania.org/archive/sth/diskimages/manifest.json   1608 entries
https://bbc.xania.org/archive/sth/tapeimages/manifest.json   1502 entries
```

Both are the exact URLs `src/sth.js` builds. Simulating `populate()` then `fetch()` end to end: the disc catalogue parses to 1,608 entries and clicking one returns a valid zip (`PK` magic, 126,633 bytes). The bracketed path this PR added encoding for — `Unreleased/Daxis[droids]-demo.zip` — fetches as `Daxis%5Bdroids%5D-demo.zip` and is byte-identical to upstream.

Manifests come back gzipped by CloudFront (125,927 bytes of JSON, 22,632 on the wire) with `access-control-allow-origin: *`, so the catalogue fetch is roughly 15x lighter than the ~350 KB of `reclist.php` HTML this replaces, and works from any origin.

No changes needed to this PR.

PR #692, opened 2026-05-11
fix: use process.defaultApp to detect packaged Electron app

### LINKED ISSUE #684: --noboot behavior inverted
When using the desktop application on Windows, --noboot actually sets autoboot to false instead of true due to the way argparse magically handles the --no prefix.

So `jsbeeb --noboot foo.dsd` runs the disk, while `jsbeeb foo.dsd` does not.

SAY-5:
Fixes #684.

In getArguments the basename(argv[0]) === "jsbeeb" check never matched on Windows (the packaged binary is jsbeeb.exe), so the function fell through to argv.slice(2) and silently dropped the first user argument. That made "jsbeeb --noboot foo.dsd" appear to autoboot (--noboot was discarded and foo.dsd became disc1) and "jsbeeb foo.dsd" appear to do nothing (foo.dsd was discarded).

Switched to process.defaultApp, which Electron sets only when running `electron .` in development, so the dev/packaged distinction is reliable across Linux, macOS, and Windows. Extracted getArguments into src/app/args.js so it can be unit-tested without booting Electron, and added tests/unit/test-app-args.js covering the Windows .exe case, the macOS .app/.../jsbeeb path, the Linux packaged path, dev mode, and the Chrome-flag filter.

### DIFF
--- a/src/app/app.js
+++ b/src/app/app.js
@@ -5,6 +5,8 @@ import * as fs from "fs";
 import * as path from "path";
 import { ArgumentParser } from "argparse";
 
+import { getArguments } from "./args.js";
+
 const store = new Store();
 
 ipcMain.on("set-title", (event, title) => {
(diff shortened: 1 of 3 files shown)


mattgodbolt (approved):
Very cool! thank you!

PR #696, opened 2026-05-12
fix: use clientX/Y for mouse coordinates

### LINKED ISSUE #631: fix: CRT mouse coordinate calculation may be incorrect with display filters
The mouse coordinate calculation in `onCubMouseEvent` mixes `evt.offsetX`/`offsetY` (relative to event target) with container bounding-rect positions. When the canvas is inset within the monitor image (e.g. display filters with non-zero canvasLeft/canvasTop), this could produce incorrect coordinates.

Consider computing x/y from `evt.clientX`/`evt.clientY` against the screen canvas rect directly:
```js
const screenRect = screenCanvas.getBoundingClientRect();
const x = (evt.clientX - screenRect.left) / screenRect.width;
const y = (evt.clientY - screenRect.top) / screenRect.height;
```

This is a pre-existing issue (present in the original jQuery code), not a regression from the jQuery removal.

Spotted by Copilot in #630.

kimjune01:
Fixes #631.

Mouse coordinates were calculated incorrectly when the canvas was inset within the monitor image (e.g., with display filters that add non-zero canvasLeft/canvasTop). The previous approach mixed evt.offsetX/offsetY with bounding rect positions, which failed in these scenarios.

Switched to calculating mouse coordinates using evt.clientX/clientY against the screen canvas rect directly. This correctly handles all display filter scenarios and eliminates the incorrect mixing of coordinate systems.

Includes a regression test that reproduces the original issue.

### DIFF
--- a/tests/unit/test-mouse-coordinates.js
+++ b/tests/unit/test-mouse-coordinates.js
@@ -1,0 +1,132 @@
+import { describe, expect, it } from "vitest";
+
+/**
+ * Test for mouse coordinate calculation logic.
+ *
+ * Issue #631: The mouse coordinate calculation in onCubMouseEvent mixes
+ * evt.offsetX/offsetY (relative to event target) with container bounding-rect
+ * positions. When the canvas is inset within the monitor image (e.g. display
+ * filters with non-zero canvasLeft/canvasTop), this produces incorrect coordinates.
+ *
+ * The correct approach is to use evt.clientX/clientY against the screen canvas
+ * rect directly.
+ */
+
+/**
+ * Simulates the CURRENT (buggy) coordinate calculation from main.js lines 541-542
+ */
+function calculateCoordinatesBuggy(evt, cubRect, screenRect, screenCanvas) {
+    const x = (evt.offsetX - cubRect.left + screenRect.left) / screenCanvas.offsetWidth;
+    const y = (evt.offsetY - cubRect.top + screenRect.top) / screenCanvas.offsetHeight;
+    return { x, y };
+}
+
+/**
+ * Simulates the CORRECT coordinate calculation using clientX/Y
+ */
+function calculateCoordinatesCorrect(evt, screenRect) {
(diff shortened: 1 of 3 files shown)


tests/unit/test-mouse-coordinates.js
```diff
@@ -0,0 +1,132 @@
...
+ * Issue #631: The mouse coordinate calculation in onCubMouseEvent mixes
+ * evt.offsetX/offsetY (relative to event target) with container bounding-rect
+ * positions. When the canvas is inset within the monitor image (e.g. display
+ * filters with non-zero canvasLeft/canvasTop), this produces incorrect coordinates.
+ *
+ * The correct approach is to use evt.clientX/clientY against the screen canvas
+ * rect directly.
+ */
+
+/**
```
mattgodbolt: No need to test the old broken implementation, can you remove this and the tests. I'm happy that the new code works :)

tests/unit/test-mouse-coordinates.js
```diff
@@ -0,0 +1,132 @@
...
+}
+
+/**
+ * Simulates the CORRECT coordinate calculation using clientX/Y
+ */
+function calculateCoordinatesCorrect(evt, screenRect) {
+    const x = (evt.clientX - screenRect.left) / screenRect.width;
+    const y = (evt.clientY - screenRect.top) / screenRect.height;
+    return { x, y };
+}
```
mattgodbolt: Can you extract this function and use it in `onCubMouseEvent` so we are testing the code we are using instead of a copy of it?

mattgodbolt (commented):
Thank you! And for the lovely tests - just a couple tweaks, I don't think we should test the old behaviour though I appreciate you demonstrating it was broken in test form; and I think we should test the code we execute and not a copy paste of it. Thnk you!

kimjune01:
OK removed old tests && extracted shared module.

mattgodbolt (approved):
Thanks!

PR #721, opened 2026-07-26
fix: defer the restart instead of discarding the change

### LINKED ISSUE #717: Declining the "restart now?" prompt still applies the change: URL is rewritten, and Electron persists it to disk
Declining the "Changing model requires a restart of the emulator. Restart now?" prompt does not decline the change. The URL is rewritten regardless, and in the Electron app the setting is written to disk regardless, so the next launch boots with the setting the user refused.

Found while reviewing #714. Entirely pre-existing; #714 neither causes nor fixes it.

### Browser

1. Load the emulator with no co-processor, e.g. `http://localhost:5173/`.
2. Open Settings, tick **65C02 co-processor**, close the dialog.
3. The "Restart now?" prompt appears. Click **No, thanks** (or press Esc, or click outside; they are all the same thing).

Observed:
- The address bar already reads `?coProcessor`, and did so before the prompt was answered.
- The running machine has no co-processor, correctly.
- Reopen Settings: the 65C02 box is **un-ticked**, contradicting the URL.

Expected: declining leaves the URL, the checkbox and the machine agreeing that there is no co-processor.

Mitigation that exists by accident: `updateUrl()` uses `pushState`, so pressing Back restores the previous URL and quietly puts things back in agreement.

### Electron (the more serious case)

1. Launch the desktop app with no co-processor.
2. Open Settings, tick **65C02 co-processor**, close the dialog.
3. Click **No, thanks**.
4. Quit and relaunch.

Observed: the app boots **with** the co-processor attached.

Expected: declining the restart means the setting was not applied, so a later launch should be unaffected.

The same applies to Econet, Music 5000, teletext and model changes, which are the other settings gated behind this prompt.

## Why

Three separate things, none of which is conditional on the user's answer.

**`areYouSure` is not a question, it is an announcement.** It wires a handler to `.ays-yes` only, then shows the modal and returns:

```js
function areYouSure(message, yesText, noText, yesFunc) {
    ...
    aysEl.querySelector(".ays-yes").addEventListener("click", function () { aysModal.hide(); yesFunc(); }, { once: true });
    aysModal.show();
}
```

There is no `.ays-no` handler at all (the button is a plain `data-bs-dismiss="modal"`), and no callback for dismissal. So declining runs no code, and `areYouSure` has already returned by the time the user sees the dialog.

**`onClose` therefore commits the change before the answer.** In `main.js`:

```js
function onClose(changed) {
    parsedQuery = Object.assign(parsedQuery, changed);   // committed immediately
    if (changed.model || changed.coProcessor !== undefined || ...) {
        areYouSure("Changing model requires a restart...", ..., function () { updateUrl(); window.location.reload(); });
    }
    ...
    updateUrl();                                          // runs synchronously, before the user answers
}
```

**Electron persists on modal close, not on confirmation.** `config.js` dispatches unconditionally when anything changed:

```js
configuration.addEventListener("hide.bs.modal", () => {
    this.onClose(this.changed);
    if (Object.keys(this.changed).length > 0) {
        this.dispatchEvent(new CustomEvent("settings-changed", { detail: this.changed }));
    }
});
```

and `src/app/electron.js` forwards it straight to disk:

```js
config.addEventListener("settings-changed", (e) => { api.saveSettings(e.detail); });
```

which reaches `src/app/app.js`:

```js
ipcMain.on("save-settings", (event, settings) => {
    const current = store.get("settings", {});
    store.set("settings", { ...current, ...settings });
});
```

Meanwhile the `Config` instance fields (`coProcessor`, `hasEconet`, …) are never updated by the checkbox listeners, which write only to `this.changed`. That is why reopening the dialog un-ticks the box: the `show.bs.modal` handler re-renders from the fields, which still hold the startup values.

## Severity

Nothing crashes and nothing is lost. The browser case is arguably defensible as "your preference is saved for next time", just never communicated, and Back undoes it. The Electron case is a straightforward correctness bug: the user was asked a yes/no question and the "no" answer is discarded.

## Fixing it is not a one-liner

`changed` mixes two kinds of setting:

- **apply-live**: `keyLayout`, `tubeCpuMultiplier`, `speechOutput`, `mouseJoystickEnabled`, `microphoneChannel`, `displayMode`. These should be applied and persisted immediately on close, as they are today.
- **needs-restart**: `model`, `coProcessor`, `hasEconet`, `hasMusic5000`, `hasTeletextAdaptor`. These should be held until confirmed.

So the shape of a fix is to split the delta, apply and persist the live half on close, hold the restart-required half until the `yes` callback fires (and only then assign into `parsedQuery`, call `updateUrl()` and dispatch `settings-changed`), and give the dialog a dismiss handler that discards the held half and re-renders the checkboxes from the `Config` fields so the UI agrees again.

That is close enough to the general `Config` state/UI restructure that it may want doing at the same time; see the discussion in #711. The Electron persistence half could reasonably be fixed on its own first, since it is the part that outlives the page.

mattgodbolt:
Declining the "Restart now?" prompt used to discard the change entirely, and
before that it applied the change but never said so. Neither matches the
question being asked: it is about *when* to restart, not *whether* to keep
the setting.

Now both answers keep the change. "Restart now" reloads; "Later" leaves the
running machine alone and the change takes effect on the next reload, or the
next launch in the Electron app. Dismissing the dialog by any route means
Later, which is the non-destructive answer.

The half of #717 that stays fixed is the contradiction: the checkbox used to
un-tick itself after you declined, disagreeing with what had actually been
saved. It now shows the saved value, with a note saying it is waiting on a
restart. The note tracks a comparison against what the running machine was
built with, so reverting a change clears it as you make it.

Two things follow from settings now being able to run ahead of the machine:

- The prompt only appears when the settings really do differ from what is
  running. Putting a setting back to what the machine already has saves it
  without offering a pointless restart.
- The model name in the top bar, and so the Electron window title, keeps
  naming the machine that is running rather than the one you have chosen.
  The pending note in the dialog is where the choice is shown.

## Also in here

Reworking this exposed how much each setting costs: a field, a click handler,
a setter that poked both the DOM and the field, and an entry in the
restart-required list, with the names lining up in none of them. Two setters
had quietly never set their fields at all.

The checkbox settings are now one descriptor list, and the handlers,
rendering, adoption and restart-required set all derive from it. The restart
decisions fall out as `needsRestart()` and `restartPending()`, pure functions
over plain objects, which is what lets them be tested in node without jsdom
or a copy of `index.html`.

Also fixes a hazard the new semantics introduce: `syncLights` polled
`processor.econet` based on the *settings*, which can now be ahead of the
machine. It asks the machine instead.

## Testing

642 unit tests pass; lint and prettier clean.

Exercised by hand in a browser, on the built app:

- Ticking the co-processor saves it, offers the prompt, and "Later" leaves the
  page running with the change in the URL. Reopening shows the box ticked and
  the note.
- Reverting a pending change clears the note as you click, and closing then
  offers no restart. Model changes behave the same way.
- With a restart pending, changing only a live setting saves it without
  asking about a restart again.
- The tube-multiplier slider follows the co-processor checkbox throughout,
  and the top bar keeps naming the running machine after declining.

Not verified: the Electron quit-and-relaunch path. `settings-changed` reaches
`api.saveSettings` as before, but no desktop build was run.

Fixes #717


mattgodbolt:
> areYouSure can accumulate multiple active click/hidden handlers if it is invoked again before the previous modal instance has finished hiding.

Correct in the abstract, but not reachable here, so I've left it alone rather than adding a guard.

`areYouSure` has exactly one call site, `onRestartRequired`, which only runs from the settings dialog's `hide.bs.modal`. To overlap two invocations you would have to reopen and re-close the settings dialog inside the ~300ms hide animation, from behind a modal backdrop. And the only `yesFunc` is `window.location.reload()`, which is idempotent, so even the overlapping case would just reload once.

Both listeners are already `{ once: true }`, and the `hidden` handler removes the click handler, so nothing leaks across invocations either.

Worth revisiting if `areYouSure` ever gains a second call site or a `yesFunc` with side effects that matter.

PR #742, opened 2026-07-30
fix: report teletext channel load failures to the user

### LINKED ISSUE #737: Teletext stream load failures are silent
`TeletextAdaptor.loadChannelStream` does not handle a failed `utils.loadData`, so a missing or unreachable `teletext/txtN.dat` surfaces only as an unhandled rejection in the console and the adaptor keeps showing whatever it last had.

The pattern for reporting this to the user already exists: `Keyboard` extends `EventTarget` and dispatches a `showError` CustomEvent (`src/keyboard.js:320`) which `main.js` bridges to the error modal (`src/main.js:778`). The adaptor cannot do the same as it stands, because it is constructed inside `Cpu6502.reset()` (`src/6502.js:1367`) and rebuilt on every hard reset, so there is nowhere stable for `main.js` to attach a listener. Fixing the reporting probably means giving the adaptor a lifetime that matches the machine rather than the reset.

Noticed while fixing #727; deliberately left out of that change.

mattgodbolt:
A failed fetch of `teletext/txtN.dat` left the adaptor showing whatever it last had, with nothing but an unhandled rejection to say so. It now dispatches a `showError` event that `main.js` bridges to the error modal, the same route `Keyboard` already uses.

The obstacle #737 describes is the adaptor's lifetime: `resetPeripherals()` rebuilt it on every hard reset, so there was nowhere stable for `main.js` to attach a listener. It is now constructed once in the `Cpu6502` constructor and `reset(true)` clears its state via a new `clearState()`, which is field-for-field what constructing a fresh one used to do. `streamRequest` is deliberately excluded so the out-of-order guard from #734 still rejects a fetch that was in flight across the reset; there is a test pinning that.

The last commit is a separate, pre-existing bug found while reviewing this one, and is easy to drop if you'd rather it went elsewhere: only a register access clears the adaptor's IRQ, so a hard reset taken while INT was latched left bit 5 of `cpu.interrupt` asserted with `teletextInts` false, and nothing would ever claim it. `clearState()` now clears it.

Fixes #737

### DIFF
--- a/src/6502.js
+++ b/src/6502.js
@@ -1425,8 +1425,7 @@ export class Cpu6502 extends Base6502 {
     buildPolltime() {
         const nop = (_cycles) => {};
         const tubeStuff = this.hasTube ? (cycles) => this.tube.execute(cycles) : nop;
-        // The adaptor itself does not exist until reset() runs, so branch on the fitting, not the object.
-        const teletextStuff = this.hasTeletextAdaptor ? (cycles) => this.teletextAdaptor.polltime(cycles) : nop;
+        const teletextStuff = this.teletextAdaptor ? (cycles) => this.teletextAdaptor.polltime(cycles) : nop;
         const musicStuff = this.music5000 ? (cycles) => this.music5000.polltime(cycles) : nop;
         const econetStuff = this.econet
             ? (cycles) => {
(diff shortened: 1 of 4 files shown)


mattgodbolt (commented):
Really unlikely to ever be a problem but hey you never know

PR #743, opened 2026-07-30
feat: offer second processor speeds that are not multiples of 3MHz

### LINKED ISSUE #733: Let the second processor run at speeds that are not multiples of 3MHz
Since #732, `tubeCpuMultiplier` is a whole-number multiple of the co-processor's own 3MHz clock, so the reachable speeds are 3, 6, 9MHz and so on. That drops the 4MHz of the Master Turbo's internal 65C102, which the old host-relative default of 2x happened to hit.

Nothing in the engine requires an integer: `Tube6502.execute` already accumulates fractional cycles. The multiplier is integral only because `ParamTypes.INT` parses the URL parameter with `parseInt` (`src/main.js:141`) and the slider has `step="1"` (`index.html`).

Suggested shape: make the control a list of real speeds (3MHz external, 4MHz Turbo, then 6/8/12/24 for #134) and parse the parameter as a float, so existing integer links keep their meaning. Below about 2.2MHz the MOS's R3 transfers lose data, so that is the floor.

mattgodbolt:
Fixes #733.

The Tube CPU slider now indexes a list of real speeds (3, 4, 6, 8, 12, 24MHz) rather than whole multiples of the parasite's 3MHz clock, so the Master Turbo's 4MHz 65C102 is reachable again. `tubeCpuMultiplier` is parsed as a float, so existing integer links mean exactly what they did.

The engine needed no change: `Tube6502.execute` accumulates in float and `polltime` only subtracts integers, so the part cycles carry over rather than being lost. There are now tests for that.

Decisions worth a look:

- The multipliers are rounded to four places, so 4MHz shares as `tubeCpuMultiplier=1.3333` rather than seventeen digits of 4/3. That runs the parasite at 3.9999MHz.
- The label leads with the speed and puts the multiplier after: `4MHz (1.33x)`, `24MHz (8x)`. Every offered speed is shorter than the `8x (24MHz)` it replaces, so it stays on one line.
- A multiplier from a URL that is not on the list is labelled as it stands (`0.75` shows `2.25MHz (0.75x)`) and the slider snaps to the nearest speed it does offer.
- The slider's `max` comes from the speed list in `config.js` so the two cannot drift apart; `aria-valuetext` carries the label now that the value itself is just a position.

https://claude.ai/code/session_01DPt4XEKzMWCccMQx5rq7uV

### DIFF
--- a/README.md
+++ b/README.md
@@ -170,8 +170,9 @@ sudo rpm -i out/dist/jsbeeb-1.0.1.x86_64.rpm
 - `cpuMultiplier=X` speeds up the CPU by a factor of `X` relative to the peripherals: video, sound and the VIAs keep
   running at their real-world rates. May be fractional or below one to slow the CPU down. NB disc loads become
   unreliable with a too-slow CPU, and running too fast might cause the browser to hang.
-- `tubeCpuMultiplier=X` speeds up the 65c02 second processor by a factor of `X`, a whole number. `1`, the default, is
-  the real 3MHz part.
+- `tubeCpuMultiplier=X` speeds up the 65c02 second processor by a factor of `X`, which may be fractional. `1`, the
+  default, is the real 3MHz part; `1.3333` is the Master Turbo's 4MHz 65C102. Below about `0.73` (2.2MHz) the MOS's
+  unhandshaken tube transfers lose data.
 - `sbLeft` / `sbRight` / `sbBottom` - a URL to place left of, right of, or below the cub monitor. The left and right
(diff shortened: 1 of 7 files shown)


mattgodbolt:
Superseded by #747, which fixes this the way you framed it: the model carries the real part's clock, so 1x is always the machine as sold, and the multiplier means overclocking alone. A Master gets the 4MHz 65C102 Turbo board, so 4MHz needs no fractional multiplier at all. Fractional multipliers still work, for overclocks that are not neat doublings.

Issue #746, opened 2026-08-02
Second processors run a CMOS 65C02 core whatever part they really had


mattgodbolt:
Both tube models in `src/models.js` use `CpuModel.CMOS65C02`, which is the Rockwell/WDC part with `RMB`/`SMB`/`BBR`/`BBS`. Neither real co-processor had that CPU:

- The **6502 Second Processor** (external, 3MHz, `6502 TUBE 1.10`) had an NMOS 6502A, with no CMOS instructions at all.
- The **65C102 Co-Processor** (the Master Turbo, 4MHz, `65C102 TUBE 1.20`) had Acorn's 65C102, which matches the `CMOS65C12` table jsbeeb already has for the Master's own CPU: the CMOS additions, but not the Rockwell bit instructions.

A superset core only ever runs more software than the real machine, so nothing is broken today. It does mean jsbeeb executes code a real second processor would reject, and the disassembler offers instructions the part never had.

Both tables already exist as `opcodes6502` and `opcodes65c12` in `src/6502.opcodes.js`, so this is largely a matter of pointing each tube model at the right one and seeing what breaks. Worth checking the parasite ROMs themselves still run, since they were written for the real parts.

PR #747, opened 2026-08-02
feat: fit each machine the second processor it was sold with

### LINKED ISSUE #733: Let the second processor run at speeds that are not multiples of 3MHz

mattgodbolt:
The tube CPU multiplier was a multiple of a hardcoded 3MHz, so a Master Turbo's 4MHz needed `tubeCpuMultiplier=1.3333`. That conflated two things in one number: which machine this is, and how much the user wants to cheat.

- `Model` carries the parasite's own `clockMhz`, so 1x always means the real part. A `tubeModel()` factory refuses a co-processor without one, since a missing clock would otherwise leave the parasite owed NaN cycles and silently never execute.
- Adds the **65C102 Co-Processor** (4MHz, `65C102 TUBE 1.20`) beside the external **6502 Second Processor** (3MHz, `6502 TUBE 1.10`). They are different parts with different parasite ROMs: 1581 of 2048 bytes differ.
- `tubeModelFor()` fits a Master with the Turbo board and everything else with the external box, and both `main.js` and `fake6502.js` go through it so the browser and the tests build the same machine. `?model=MasterTurbo` is therefore now genuinely one.
- The multiplier means overclocking alone and may be fractional, since a 60% overclock is no less reasonable than a doubling. The slider steps in 0.05 and the label follows the fitted tube: `1x (4MHz)` on a Master, `1.6x (4.8MHz)` boosted.

Supersedes #743, which reached 4MHz by offering a fixed list of speeds including 1.3333x.

Fixes #733

## Verification

`tests/integration/tube.js` builds a Master with a tube through `fake6502`, so the boot regression test from #732 now runs the new ROM at 4MHz and still reaches the BASIC prompt across the tube. New unit tests pin the clock each machine gets and the NOPs it retires per host cycle at 1x and 1.6x, which is what catches truncated fractional accumulation.

A review pass measured 120 emulated seconds at 1.05x, 1.6x and 4/3x: the cycle accumulator never exceeded 2.0 and residue after two emulated minutes was 3.6e-8 cycles, so fractional multipliers do not drift.

## Worth knowing

- **`?model=Master&coProcessor` now runs at 4MHz where it ran at 3MHz.** That is the second re-reading of this parameter after #732, and it is deliberate: a Master's co-processor was the Turbo board. It does mean an external box on a Master is no longer expressible, which wants a tube choice in the config dialog rather than a multiplier.
- A range input snaps a programmatically set value to its step, so `?tubeCpuMultiplier=1.3333` leaves the knob at 1.35 while the machine runs 1.3333 and the label reads the true value. Pre-existing for out-of-range values; fractional support makes it reachable by design.
- `TODO(#746)` on both models: neither real part was a Rockwell 65C02. The external box had an NMOS 6502A and the Turbo had Acorn's 65C102, which matches the `CMOS65C12` table already in the tree.


src/6502.js
```diff
@@ -15,10 +15,8 @@ import { AtomMMC2 } from "./mmc.js";
 
 const signExtend = utils.signExtend;
 
-export const TubeCpuClockMhz = 3;
 const HostClockMhz = 2;
-const TubeCyclesPerHostCycle = TubeCpuClockMhz / HostClockMhz;
-// Below about 2.2MHz the MOS's unhandshaken 10us/byte R3 transfers lose data.
+// A multiple of the fitted co-processor's own clock, so 1x is the real part.
```
mattgodbolt: kill this pointless constant and comment. A default of `1` is fine to inline

src/6502.js
```diff
@@ -416,7 +414,12 @@ class Tube6502 extends Base6502 {
     constructor(model, cpu, { cpuMultiplier = DefaultTubeCpuMultiplier } = {}) {
         super(model, { cycleAccurate: false });
 
+        // A negative multiplier leaves the parasite owed ever fewer cycles, so it stops executing
```
mattgodbolt: kill this pointless comment. of course a negative parameter is dumb, the message thrown is enough

src/models.js
```diff
@@ -65,6 +66,12 @@ function pickDfs(cmos) {
     return cmos;
 }
 
+/** A second processor. Its clock is what the multiplier in the emulation config multiplies. */
+function tubeModel({ name, os, cpuModel, clockMhz }) {
+    if (!(clockMhz > 0)) throw new Error(`Co-processor ${name} has no clock speed`);
```
mattgodbolt: why the hell is this not in the Model constructor? We shouldn't have a "clockMhz" that is nullable/Nanable in the Model.,

src/6502.js
mattgodbolt: Gone, inlined as `{ cpuMultiplier = 1 }` in the constructor, along with its comment. `main.js` and the test now say 1 too.

src/6502.js
mattgodbolt: Comment gone; the throw stands on its own.

src/models.js
mattgodbolt: Agreed, and you are right that it was incongruous. Every model now declares `clockMhz` (2 for the Beebs, 1 for the Atom, 3 and 4 for the two co-processors) and the `Model` constructor throws without one, so the factory is gone and the field is neither nullable nor NaN-able. `Tube6502` now divides by `cpu.model.clockMhz` rather than a local `HostClockMhz`, and `main.js` takes `cpuSpeed` from the model instead of branching on `isAtom`. The remaining hardcoded 2MHz sites are filed as #751 rather than fixed here.

Issue #748, opened 2026-08-02
Wish: remap keys from within the emulator, not just via URL parameters


mattgodbolt:
Keys can be remapped today, but only by hand-editing the URL:

```
https://bbc.xania.org/?disc1=sth:Superior/SpaceInvaders-Superior.zip&autoboot&KEY.ENTER=COPY
```

That works, but it asks a lot of the person who most needs it: they have to know the parameter exists, know that the
host-side name for Return is `ENTER` while the BBC-side name is `RETURN`, and get it right without any feedback beyond
the game not responding. (A dialog now reports names it doesn't recognise, which helps with typos but not with
discovery.)

Plenty of games from the era chose keys that are awkward or impossible on a modern keyboard, particularly on a Mac:

- `COPY` is `End`, ie `fn`+`→`
- `CAPS LOCK` toggles rather than acting as a key you hold, so games that steer with it are unplayable as-is

It would be much friendlier to do this in the emulator itself. Roughly:

- somewhere in the UI (Settings? a "Keys" menu?) showing the current mapping
- "click a BBC key, then press the key you want to use for it", so nobody has to learn either set of names
- reflect the result in the URL, so a working setup can be bookmarked and shared, which is the whole appeal of the
  current parameters
- ideally remember it per disc image, since the awkward keys are a property of the game

Worth considering at the same time:

- the D-pad's default gamepad mapping has no name, so it can't be remapped at all (see `GP.` parameters)
- the Atom uses a different set of key names again

Prompted by a note from one of the original Superior Software authors, who found their own games hard to play now
because of the keys they chose in 1984.

PR #750, opened 2026-08-02
feat: document key remapping, and report mappings that can't be applied

mattgodbolt:
Prompted by a note from one of the original Superior Software authors, who can't comfortably play his own games any
more: Invaders fires with `COPY`, Hunchback steers with `CAPS LOCK` and `CTRL`. The `KEY.` URL parameters have always
been able to fix that, but they were undocumented and failed silently.

Documents them, and fixes what didn't work.

### Documentation

A "Remapping Keys" section in the README with the syntax, the host-side and BBC-side name lists, and three worked
examples, all of which load and autoboot today:

- [`?disc1=sth:Superior/SpaceInvaders-Superior.zip&autoboot&KEY.ENTER=COPY`](https://bbc.xania.org/?disc1=sth%3ASuperior%2FSpaceInvaders-Superior.zip&autoboot&KEY.ENTER=COPY)
- [`?disc1=sth:Superior/Frogger-Superior.zip&autoboot&KEY.UP=A&KEY.DOWN=Z&KEY.LEFT=DELETE&KEY.RIGHT=COPY`](https://bbc.xania.org/?disc1=sth%3ASuperior%2FFrogger-Superior.zip&autoboot&KEY.UP=A&KEY.DOWN=Z&KEY.LEFT=DELETE&KEY.RIGHT=COPY)
- [`?disc1=sth:Superior/Hunchback-Superior.zip&autoboot&KEY.LEFT=CAPSLOCK&KEY.RIGHT=CTRL`](https://bbc.xania.org/?disc1=sth%3ASuperior%2FHunchback-Superior.zip&autoboot&KEY.LEFT=CAPSLOCK&KEY.RIGHT=CTRL)

The `GP.` gamepad parameters are documented too, in the Joystick Support section.

### Behaviour

Mappings that can't be applied now appear in the error dialog rather than only in the console, and say which half was
wrong. `KEY.RETURN=COPY` is the obvious trap — the key the BBC calls `RETURN` is `ENTER` on the host side — and
previously gave no feedback at all.

### Fixes

- **The mapping was applied once, then lost.** `userKeymap` was drained with `pop()` while building the key map, so
  changing keyboard layout or model silently reverted it.
- **`KEY.` did nothing on the Atom.** `getKeyMapAtom` read `mapping.atom`; `processKeyboardParams` only ever wrote
  `mapping.bbc`. Entries are now `{native, key}`, validated against `ATOM` or `BBC` depending on the model — which
  means processing them after the model is resolved.
- **`GP.x=0` was rejected**, as `"0123456789".indexOf(key) > 0` excludes `"0"` itself.

Also adds a `remap()` helper so a deliberate override no longer logs `Warning: duplicate binding for key 13`.

Tests cover the override applying in both shift states, surviving a rebuild, working on the Atom, unknown names being
reported, and `GamePad.remap`'s new return contract.

Not addressed here, and now filed as #748: doing any of this from the UI instead of the URL. The D-pad's default
gamepad mapping also has no name, so it can't be remapped at all; noted in the README.

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -52,7 +52,7 @@ import {
     parseMediaParams,
     parseQueryString,
     processAutobootParams,
-    processKeyboardParams,
+    processInputParams,
 } from "./url-params.js";
 
 let processor;
@@ -297,10 +297,8 @@ config.setDisplayMode(displayMode);
 
 model = config.model;
 
-// Process keyboard and gamepad mappings. Must come after the model is known: `KEY.`
-// parameters are validated against the emulated machine's key names, which differ on the
-// Atom. Anything we couldn't apply is shown below, once the error dialog is available.
-const keyMappingWarnings = processKeyboardParams(
+// Must come after we know the model, to validate names against those of the hardware.
+const keyMappingWarnings = processInputParams(
     parsedQuery,
     model.isAtom ? utils_atom.ATOM : BBC,
     keyCodes,
@@ -383,8 +381,6 @@ function showError(context, error) {
     errorDialogModal.show();
 }
 
-// A mistyped `KEY.`/`GP.` parameter is easy to miss otherwise: the emulator starts up
-// looking perfectly normal, but with the keys still where the game put them.
 if (keyMappingWarnings.length) {
     showError("applying the key mappings in the URL", keyMappingWarnings.join(" "));
 }
--- a/src/main.js
+++ b/src/main.js
@@ -177,9 +177,6 @@ const { discImage: queryDiscImage, secondDiscImage: querySecondDisc, mmcImage } 
 if (queryDiscImage) discImage = queryDiscImage;
 if (querySecondDisc) secondDiscImage = querySecondDisc;
 
-// Process keyboard mappings
-parsedQuery = processKeyboardParams(parsedQuery, BBC, keyCodes, utils.userKeymap, gamepad);
-
 // Handle specific query parameters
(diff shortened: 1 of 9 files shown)


src/main.js
```diff
@@ -375,6 +383,12 @@ function showError(context, error) {
     errorDialogModal.show();
 }
 
+// A mistyped `KEY.`/`GP.` parameter is easy to miss otherwise: the emulator starts up
```
mattgodbolt: Please remove the pointless comment, we can see that's what it's doing.

src/main.js
```diff
@@ -300,6 +297,17 @@ config.setDisplayMode(displayMode);
 
 model = config.model;
 
+// Process keyboard and gamepad mappings. Must come after the model is known: `KEY.`
```
mattgodbolt: Please remove the inane commentary. Only "Keyboard mappings must be processed after we know the model to validate its names against those of the hardware" or similar is relevant here, the rest is editorialising and/or obvious from the variable names

src/main.js
```diff
@@ -300,6 +297,17 @@ config.setDisplayMode(displayMode);
...
+// parameters are validated against the emulated machine's key names, which differ on the
+// Atom. Anything we couldn't apply is shown below, once the error dialog is available.
+const keyMappingWarnings = processKeyboardParams(
```
mattgodbolt: please rename this function as it's no longer "keyboard parameters". that also obviates the comment

PR #753, opened 2026-08-02
perf: decode flux marks with 32 bit words instead of BigInts

### LINKED ISSUE #749: Sector scanning decodes flux bit by bit through BigInts, and it is slow
Decoding a disc's sectors from its flux is slow enough to be felt: it is why the two Elite HFE tests needed timeout overrides in #744, and why a save or a catalogue read of an HFE pauses noticeably.

## Where the time goes

`Track.findSectorIds` (`src/disc.js:388`) re-derives the sector layout from the raw pulses by walking the track one bit at a time looking for FM and MFM address marks:

```js
const bitLength = this.length * 32;
for (let pulseIndex = 0; pulseIndex < bitLength; ++pulseIndex) {
```

For the bundled `public/discs/elite.hfe`, each track side is 3128 words, so 100,096 iterations per side and **16,208,576 for a full 162-side disc**. Measured on a dev machine:

| | |
| --- | --- |
| one side | 26ms |
| full disc scan | ~2.0s |

The mark detector needs a 64-bit sliding window, so `markDetector` and `markDetectorPrev` are `BigInt`s shifted and masked on every bit. That dominates. The same loop shape, 16.2M iterations, shift + mask + compare:

| representation | time |
| --- | --- |
| `BigInt` (what the loop does today) | 1302ms |
| `Number` | 90ms |

So roughly 14x, which accounts for most of the two seconds.

## What is worth trying

The obvious idea is to hold the 64-bit window as a pair of 32-bit Numbers. Worth saying plainly that this is speculation, not a plan: the original C this was transliterated from used 64-bit integers throughout, and the mark detection compares against 64-bit patterns (`0xaaaa448944894489n` for MFM), so a split representation has to keep those comparisons exactly right. Getting it subtly wrong means misreading protected discs, which is precisely the class of bug that is hardest to notice.

Other angles, cheaper or complementary:

- Cache the decode per track and invalidate on write. Callers scan the same unmodified tracks repeatedly: `ssdOrDsdShortfalls` and `toSsdOrDsd` each do a full pass, and the catalogue reader does another.
- Skip tracks with no pulses at all before entering the bit loop.
- Only decode sector data when the caller needs it. `findSectors` always calls `Sector.read()`, but `ssdOrDsdShortfalls` only needs headers for most of its checks.

Any change here needs the existing HFE round-trip tests plus a check against the real captures used in #740: Elite, Carousel and Stairway to Hell all contain things a clean image does not, and are the discs most likely to expose a decode regression.

mattgodbolt:
The 64 bit mark detector in `Track.findSectorIds` and the 64 bits behind it were `BigInt`s, shifted and masked once per bit: 100,096 times per side, 16.2M for a double-sided 81 track disc. Both are now `BitWindow64`, a 64 bit shift register held as a pair of unsigned 32 bit halves.

The comparisons are the same ones. FM matches the high half only (`0x88888888`, the tail of the zero sync run); MFM matches both halves of `0xaaaa448944894489`. The sync run count used to walk one BigInt a nibble at a time and now walks the low half, then the high half only if the low half matched all the way up, which is exactly when the old loop would have carried into bits 32-63.

## Measurements

Node 24, this machine. Elite is the bundled `public/discs/elite.hfe` (81 tracks, double sided); Stairway to Hell is the #740 capture (81 tracks, single sided).

`findSectorIds` alone, over all 162 sides of elite, best of seven, before and after interleaved to keep machine drift out of it:

| | |
| --- | --- |
| before | 1695ms |
| after | ~115ms |

Full disc scan via `ssdOrDsdShortfalls`, which also reads every sector's data:

| | before | after |
| --- | --- | --- |
| elite | 1960ms | ~150ms |
| Stairway to Hell | 1214ms | ~85ms |

Holding the halves in a class rather than four flat locals costs about 25% of the bit loop, ~115ms against ~90ms over 162 sides. That is the deliberate price of having the shifting and the run count behind an API with unit tests, in a decoder whose failure mode is silently misreading a protected disc.

## Correctness

The output is byte-identical on all five real captures from #740. For each disc I hashed every decoded sector (MFM flag, id and data bit offsets, deleted flag, both CRC error flags, byte length, header bytes, sector data) and compared before and after:

| disc | digest before == after |
| --- | --- |
| Elite (bundled) | yes |
| Carousel | yes |
| Stairway to Hell | yes |
| Perplexity | yes |
| Disc Drive Utilities | yes |

Also differential-fuzzed the old and new implementations against each other over 400 randomly generated tracks, with the pulse stream biased towards the marker and sync patterns so the branches actually fire: identical sector output and identical console diagnostics, including the "Short zeros sync" warning that exercises the split nibble count.

`BitWindow64` has unit tests of its own in `test-disc.js`, driven from BigInt values: bits crossing bit 31 and bit 63, the bit leaving the top, full-width equality, and nibble runs that stop mid-half, carry across the half boundary, fill all 16 nibbles, or are empty.

Unit tests pass, as does `tests/integration/protection.js`. The three integration files that fail here fail on `main` too, for a missing `dp111_6502Timing` submodule.

## What this does not do

The issue lists three cheaper wins alongside this one. None are here:

- **Per-track caching with invalidation on write.** `toSsdOrDsd` scans every track twice and the catalogue reader scans again, so it would buy roughly another 2x. But `pulses2Us` is written directly by `TrackBuilder`, `Disc.writePulses`, `Disc.restoreState` and `disc-hfe.js`, and a missed invalidation means silently serving a stale sector layout. That is a worse failure than a slow scan, and 150ms is no longer a scan anyone waits on.
- **Skipping tracks with no pulses.** Worth about 1ms per blank side now.
- **Deferring `Sector.read()` until a caller wants data.** `sectorShortfall` needs `sectorData` for two of its four checks, so `ssdOrDsdShortfalls` would still read most sectors.

## Risk that remains

This is a transliteration of 64 bit C arithmetic into 32 bit pairs, and the failure mode is a misread protected disc rather than a crash. The evidence above is five real captures and a fuzz run, not a proof. The `elite.hfe` timeout overrides added in #744 are left in place; they are now generous rather than necessary.

Fixes #749

### DIFF
--- a/src/disc.js
+++ b/src/disc.js
@@ -355,24 +355,62 @@ class Sector {
     }
 }
 
-// What the 64 bit mark detector holds at an address mark, split into high and low 32 bit halves.
-// The FM case only pins down the high half, the last eight zero bits of sync (0x88888888 being
-// FM data 0x00 with all clocks set); the low half is the marker byte itself, decoded separately.
-// The MFM case matches the whole of 0xaaaa448944894489.
+/** Nibbles in each 32 bit half of a {@link BitWindow64}. */
+const NibblesPerHalf = 8;
+
+/**
+ * A 64 bit shift register, held as a pair of unsigned 32 bit halves. BigInt would say this more
+ * directly but costs an order of magnitude more per bit shifted.
+ */
+export class BitWindow64 {
+    constructor() {
+        this.hi = 0;
+        this.lo = 0;
+    }
+
+    /**
+     * @param {Number} bit 0 or 1, shifted into bit 0
+     * @returns {Number} the bit shifted out of bit 63
+     */
+    shiftIn(bit) {
+        const shiftedOut = this.hi >>> 31;
+        this.hi = ((this.hi << 1) | (this.lo >>> 31)) >>> 0;
+        this.lo = ((this.lo << 1) | bit) >>> 0;
+        return shiftedOut;
+    }
+
+    /**
+     * @returns {boolean} whether all 64 bits match the halves given
+     */
+    equals(hi, lo) {
+        return this.hi === hi && this.lo === lo;
+    }
+
+    /**
+     * @param {Number} nibble
+     * @returns {Number} the length of the run of `nibble` ending at bit 0
+     */
+    countTrailingNibbles(nibble) {
+        let count = 0;
+        for (let bits = this.lo; (bits & 0xf) === nibble; bits >>>= 4) count++;
+        // Only a low half that matched all the way up can have a run continuing into the high half.
+        if (count === NibblesPerHalf) for (let bits = this.hi; (bits & 0xf) === nibble; bits >>>= 4) count++;
+        return count;
+    }
+}
+
+// What the mark detector holds at an address mark. The FM case only pins down the high half,
+// the tail of the zero sync run; the low half is the marker byte itself, decoded separately.
 const FmSyncHi = 0x88888888;
 const MfmMarkerHi = 0xaaaa4489;
 const MfmMarkerLo = 0x44894489;
-const NibblesPerWord = 8;
-// Sync ahead of an address mark is counted in zero data bits, so this is two bytes' worth.
-// A run no longer than this is unusual enough to be worth logging.
+// One data bit of zero, FM encoded with its clock, is the pulse nibble 0x8.
+const FmZeroBitPulses = 0x8;
+// FmSyncHi is eight of those, so a match has already seen this much of the sync run.
+const FmSyncHiZeroBits = 8;
+// Sync is counted in zero data bits, so a run no longer than two bytes' worth is short enough
+// to be worth logging.
 const ShortSyncZeros = 16;
-
-/** @returns {Number} how many low nibbles of `word` are the FM pulses for a zero data bit */
-function countSyncNibbles(word) {
-    let count = 0;
-    for (let bits = word; (bits & 0xf) === 0x8; bits >>>= 4) count++;
-    return count;
-}
 
 class Track {
     constructor(upper, trackNum, initialByte) {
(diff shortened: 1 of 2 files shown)


src/disc.js
```diff
@@ -355,6 +355,25 @@ class Sector {
...
+// The MFM case matches the whole of 0xaaaa448944894489.
+const FmSyncHi = 0x88888888;
+const MfmMarkerHi = 0xaaaa4489;
+const MfmMarkerLo = 0x44894489;
+const NibblesPerWord = 8;
+// Sync ahead of an address mark is counted in zero data bits, so this is two bytes' worth.
+// A run no longer than this is unusual enough to be worth logging.
+const ShortSyncZeros = 16;
+
+/** @returns {Number} how many low nibbles of `word` are the FM pulses for a zero data bit */
```
mattgodbolt: Isn't this "How man runs of low nibbles"?

src/disc.js
```diff
@@ -355,6 +355,25 @@ class Sector {
...
 }
 
+// What the 64 bit mark detector holds at an address mark, split into high and low 32 bit halves.
+// The FM case only pins down the high half, the last eight zero bits of sync (0x88888888 being
+// FM data 0x00 with all clocks set); the low half is the marker byte itself, decoded separately.
+// The MFM case matches the whole of 0xaaaa448944894489.
+const FmSyncHi = 0x88888888;
+const MfmMarkerHi = 0xaaaa4489;
+const MfmMarkerLo = 0x44894489;
+const NibblesPerWord = 8;
```
mattgodbolt: it's very unclear what a "word" is here. it seems it's 32 bits? is that the case? Do we need this constant? Maybe if we defined "word" to be 32-bits somewhere above this makes more sense?

src/disc.js
```diff
@@ -394,40 +413,40 @@ class Track {
...
-        const top32of64b = 0xffffffff00000000n;
-        const fmMarker = 0x8888888800000000n;
-        const mfmMarker = 0xaaaa448944894489n;
+        // The mark detector is a 64 bit sliding window over the pulse stream, and its previous
+        // 64 bits. Both are pairs of unsigned 32 bit Numbers rather than BigInts, which cost an
+        // order of magnitude more in a loop that runs once per bit of every track.
+        let markDetectorHi = 0;
+        let markDetectorLo = 0;
+        let markDetectorPrevHi = 0;
+        let markDetectorPrevLo = 0;
```
mattgodbolt: Would it make sense to make a little `class` for this and then everything hides away nicely inside of it. Unless it undoes the performance benefit.

mattgodbolt (commented):
No issues with correctness; lgtm. some commentary and maybe a testable refactoring of a 64-bit value class ?

src/disc.js
mattgodbolt: You are right, it is a run and not a total. It is now `BitWindow64.countTrailingNibbles(nibble)`, documented as "the length of the run of `nibble` ending at bit 0", and the caller passes `FmZeroBitPulses` (0x8) rather than the function knowing about FM. There are unit tests for the run stopping mid-half, carrying across the half boundary, filling all 16 nibbles, and being zero-length.

src/disc.js
mattgodbolt: Agreed, "word" was doing no work. It is `NibblesPerHalf` now, and "half" is defined by the thing it belongs to: `BitWindow64` is documented as a 64 bit shift register held as a pair of unsigned 32 bit halves, and the constant sits directly above it as "nibbles in each 32 bit half of a BitWindow64".

It is still needed: it is the test for whether a nibble run filled the low half and so may continue into the high one.

src/disc.js
mattgodbolt: Done, in 4bf4ec2. `BitWindow64` owns the shifting, the 64 bit equality test and the nibble run count; the scan loop now reads `markDetectorPrev.shiftIn(markDetector.shiftIn(pulseBit))`, which also makes the ordering self-evident rather than something you have to check statement by statement. It is exported and has its own unit tests in `test-disc.js`.

It does cost something, so here is the measurement rather than a claim. `findSectorIds` over all 162 sides of `elite.hfe`, best of seven, interleaved runs to keep machine drift out of it:

| | |
| --- | --- |
| BigInt (before this PR) | 1695ms |
| four flat locals | ~90ms |
| `BitWindow64` | ~115ms |

So about 25% of the bit loop, roughly 20ms on a full disc scan. I took the class: it does not undo the benefit, and a decoder where a subtle bug means silently misreading a protected disc is worth more when the bit twiddling is testable in isolation.

Two shapes I tried and rejected, both slower than the version pushed: one class holding all four halves with a single `shiftIn` (~162ms, V8 seems to handle the four-field object worse), and signed int32 halves to keep the values in SMIs (~equal, but it needs `| 0` on `0x88888888` and `0xaaaa4489` at the constants, which obscures exactly the patterns a reader wants to check against beebjit).

mattgodbolt:
Pushed 4bf4ec2 with all three points addressed; replies inline on each. Summary:

- `countSyncNibbles` is now `BitWindow64.countTrailingNibbles(nibble)`, documented as a run length, and no longer knows about FM.
- `NibblesPerWord` is `NibblesPerHalf`, with "half" defined by the class it belongs to.
- The pair of 32 bit halves is a `BitWindow64` class with its own unit tests. It costs about 25% of the bit loop (~115ms against ~90ms over 162 sides, versus 1695ms before); measurements and the two shapes I tried and rejected are on that thread.

No decode change: re-ran the differential fuzz and the five capture digests after the refactor, all still identical.

src/disc.js
```diff
@@ -355,6 +355,63 @@ class Sector {
     }
 }
 
+/** Nibbles in each 32 bit half of a {@link BitWindow64}. */
```
mattgodbolt: Can we not make this a constant in `BitWindow64` ? seems like something it should "know" and not a separate constant that's linked to it.

mattgodbolt: In fact, this is only used inside the class, right? Maybe define it in the function that needs it

mattgodbolt: Agreed on both, and your follow-up is the better of the two: `countTrailingNibbles` is the only thing that needs it, so it is a local `nibblesPerHalf` in that method now rather than a module constant or a class field. 9412533.

Re-ran the differential fuzz and the five capture digests after the move; all unchanged.

Issue #751, opened 2026-08-02
CPU clock speed is hardcoded in several places now the model carries it


mattgodbolt:
#747 gave every `Model` a `clockMhz`, asserted in the `Model` constructor, and used it for the second processor's cycle ratio and for `cpuSpeed` in `main.js`. Several other places still assume 2MHz (or branch on `isAtom` to pick 1MHz) rather than asking the model.

Known sites:

- `src/6502.js:669` — `this.peripheralCyclesPerSecond = 2 * 1000 * 1000;`
- `src/6502.js:1482` — the one-second cycle-counter rollover, `2 * 1000 * 1000` three times
- `src/acia.js:269` — `Math.floor(2 * 1000 * 1000 * sec)` converting seconds to cycles
- `src/tapes.js:4,32` — `BbcCpuSpeed` and `AtomCpuSpeed` constants, selected with an `isAtom` flag
- `src/econet.js:159` — a `1000000` cycle timeout
- `src/6847.js:347` — the Atom's video clock is derived assuming a 1MHz CPU

Most are on paths that take a flag or nothing at all rather than a model, so this is not a mechanical find-and-replace: `tapes.js` is handed `isAtom`, and the ACIA has no model reference. Some may also be genuinely fixed rather than machine-dependent (the ACIA's rate arithmetic is about the serial clock, not the CPU), so each wants deciding on its merits rather than a blanket substitution.

No user-visible bug today: every value matches the machine it applies to. This is about there being one answer to "how fast does this machine run" instead of several.

PR #755, opened 2026-08-02
refactor: ask the model how fast the machine runs

mattgodbolt:
Closes most of #751: the sites that assumed 2MHz, or picked 1MHz from an `isAtom` flag, now ask the model. `Model` gains a `cyclesPerSecond` getter and it is threaded to:

- `6502.js` — `peripheralCyclesPerSecond` (the Atom's separate 1MHz override goes away), the one-second cycle-counter rollover, and `TouchScreen`
- `acia.js` — `secondsToCycles`. This converts a duration into scheduler ticks, and the scheduler counts CPU cycles, so it is the CPU clock rather than the serial clock
- `tapes.js` — `loadTapeFromData` takes the model instead of an `isAtom` flag; the flag only ever selected between two clocks and the Atom waveform encoding, and the model answers both. The `TapefileTape` bit rate and its five seconds of carrier come from the model too
- `econet.js` — the retry timeout, now half a second of whatever clock
- `6847.js` — the VDG advances `3.638004 / clockMhz` cycles per CPU cycle rather than a constant that quietly assumed 1MHz
- `main.js`, `touchscreen.js`, and the test harness's seconds-to-cycles conversions

Values are identical for every machine, checked mechanically across all models: for each site, the old constant equals what the model derives on the machines that site applies to.

### The one deliberate behaviour change

Second commit, on the Atom only. The rollover subtracted a fixed 2e6 whatever the machine, but consumers rebuild absolute time as `cycleSeconds * clock + currentCycles` — `AtomSoundChip.updateSpeaker` divides by 1MHz, as does `tests/test-machine.js`. So on the Atom, emulated time jumped backwards by 1e6 every 2e6 cycles. Demonstrated by running an Atom past the rollover before the change (time went backwards once) and after (it did not). `docs/snapshot-format.md` already describes `cycleSeconds` as elapsed seconds; now it is.

Also worth knowing: the ACIA and touchscreen are constructed on the Atom too, and now derive 1MHz there. Nothing on the Atom reaches either — its address decode has no 6850 or serial ULA, so no ACIA task is ever scheduled and the touchscreen is never put into poll mode — so this is inert, but it is a difference.

### Left alone

`soundchip.js`'s `sampleRate / 2000000`. The base `SoundChip` is only ever fitted to a 2MHz machine; the Atom gets `AtomSoundChip`, which already takes the clock from the model via `main.js`. Threading a clock there means a required option at five construction sites, two of them test helpers, for no change in any number. Happy to do it if you'd rather have no exceptions.

### Tests

New tests fail if the clock is wrong: tape gap and carrier lengths on a 1MHz vs 2MHz machine, ACIA byte time, VDG cycles per CPU cycle, and the rollover on both clocks. Verified by mutating `cyclesPerSecond` to a constant 2MHz and watching five of them fail.

Lint, 682 unit tests, all integration tests (`--no-file-parallelism`), and the CPU suite pass.

Copilot spotted that `TouchScreen` schedules its poll callback unbound. That is real, but pre-existing and untouched here: this PR only changed the `PollCycles` constant next door. Tracked separately in #758.

### DIFF
--- a/src/6502.js
+++ b/src/6502.js
@@ -666,7 +666,7 @@ export class Cpu6502 extends Base6502 {
         this.resetLine = true;
         this.cpuMultiplier = this.config.cpuMultiplier;
         this.videoCyclesBatch = this.config.videoCyclesBatch | 0;
-        this.peripheralCyclesPerSecond = 2 * 1000 * 1000;
+        this.peripheralCyclesPerSecond = model.cyclesPerSecond;
         this.hasTube = !!this.config.tube;
         this.hasMusic5000 = !!this.config.hasMusic5000;
         this.hasTeletextAdaptor = !!this.config.hasTeletextAdaptor;
@@ -1361,7 +1361,7 @@ export class Cpu6502 extends Base6502 {
         this.fdc.powerOnReset();
         this.adconverter.reset();
 
-        this.touchScreen = new TouchScreen(this.scheduler);
+        this.touchScreen = new TouchScreen(this.scheduler, this.model.cyclesPerSecond);
         if (this.econet) this.filestore = new Filestore(this, this.econet);
     }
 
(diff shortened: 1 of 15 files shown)


mattgodbolt (commented):
LGTM

PR #762, opened 2026-08-02
feat: add an xBR display mode that smooths the picture

### LINKED ISSUE #75: Option to render with Hqx scaling
Looks pretty cool :)
https://en.wikipedia.org/wiki/Hqx

mattgodbolt:
Fixes #75.

Adds a third display mode, `Smoothed (xBR)`, alongside `RGB Monitor` and `PAL TV`. It runs Hyllian's xBR-lv2 edge-directed upscaler over the picture, so diagonals are drawn as lines rather than staircases. Reachable from the display menu or with `?displayMode=xbr`.

The issue asks for hqx. I went with xBR-lv2 instead: it is what most emulators moved to after hqx, needs no 256-entry lookup table, and suits the BBC's chunky low-colour modes better. Say the word if you specifically want hqx and I will do that instead — the plumbing is the reusable part.

### The part that isn't the algorithm

The algorithm is a port of a well-known shader. The interesting part is what it is applied to, and it is why [#667](https://github.com/mattgodbolt/jsbeeb/pull/667) produced no visible difference at all.

Our framebuffer is a 1024-wide **raster**, not a grid of BBC pixels. One logical pixel covers 1, 2, 4 or 8 texels horizontally depending on the ULA's colour bits, and usually two rows vertically. An upscaler that samples raw texel neighbours therefore finds nine copies of the same pixel in its 3x3 neighbourhood, concludes there is no edge anywhere, and changes nothing.

So `Video` now records one descriptor byte per framebuffer row — the grid changes mid-frame, a MODE 7 line above a MODE 1 playfield being routine — and the filter uploads that as a texture the shader reads. The descriptor holds the width less one, so it needs no logarithm to write and no exponential to read. In the hot loop it costs one or two byte stores per character cell; measured against `main` the difference is below the benchmark's noise floor (runs vary +/- 6%, the change measured 3% *faster*).

`docs/xbr-display-mode.md` has the full write-up.

### How the shader is tested

Node has no WebGL, so `tests/shader` gives the GLSL to a browser: it lays patterns into a framebuffer, draws them through the real shader in headless Chrome, reads the pixels back, and asserts on them. One Chrome launch covers every pattern in a run.

The assertions are the promises the mode makes — a flat area comes back untouched, a straight edge stays hard (which is what separates xBR from bilinear), a diagonal gains shades that lie between the two colours already there rather than inventing new ones, a lone pixel is chamfered without bleeding outside its cell.

Three of them are the jsbeeb-specific ones, and could not have existed before: the same picture drawn with MODE 1's two texels per pixel, MODE 2's four, and with doubled scanlines must render identically. That is the logical pixel grid asserted head-on, and forcing `pixelSize = vec2(1.0, 1.0)` — exactly the #667 failure mode — fails those three and nothing else.

This replaces an earlier design in which a JavaScript port of the algorithm was the reference and the shader was pixel-compared against it. That existed because the shader was believed untestable outside a browser; once it proved runnable from Node, the premise was gone. The comparison could only ever show the two agreed, never that either was right, and two implementations written from one reading of one source can be wrong together. Asserting behaviour on the artefact that ships is both stronger and about 470 lines smaller.

### What it suits

- **Graphics modes** are what it is for — diagonals and curves become continuous.
- **MODE 0** is barely touched. One texel per pixel and 640 across leaves little to reconstruct, and text stays crisp.
- **Chunky text in MODE 2 and 5** is where opinions will differ; eight-pixel-wide characters get noticeably rounded.
- **MODE 7** is left alone — the SAA5050's own character rounding got there first.

### Review found several things I had wrong

Three rounds of review, each of which found something real. The follow-up commits are the fixes.

**On the Atom, this did nothing.** The 6847 blits straight into the framebuffer and never recorded a grid, so every row read as unrendered, the shader saw nine identical texels and returned the centre pixel — the exact failure of #667, on a different machine. Its blitters now record the grid.

**Then it recorded half the width in every 2bpp Atom mode** — Chuckie Egg, Flappy Bird, Breakout, Fizzle Bricks. `blitPixels` steps its bit groups by `pixelsPerBit / bpp` but reads the colour with `j & 0xe`, so at 2bpp pairs of groups share a colour and a pixel is twice as wide. That is worse than doing nothing: the shader reconstructs edges on a stretched grid, turning a 45 degree edge into roughly 27.

**MODE 2 is four texels per logical pixel, not eight.** Eight is `ulaMode` 0, which no OS MODE selects. I had written eight in four places.

**Falling back could take the emulator with it.** A canvas that has handed out a WebGL context can never hand out a 2D one, so when a filter refused to build, `bestCanvas`'s 2D fallback threw. The shape was pre-existing, but the new precision check made it reachable by policy rather than only by driver bug. It now retries the passthrough filter on the existing context.

Two of my own tests turned out to prove less than I thought, both caught by review:

- the stripe program used `GCOL 0,X% MOD 4` while stepping by 8, so the colour was always 0 — it drew black on black and every assertion held on an empty screen;
- the 6847 test measured what the blitters wrote but supplied the width itself, so it asserted the arithmetic rather than the decision. Reverting the production line still passed everything.

Both are fixed, and both fixes are confirmed by putting the bug back and watching the right tests fail. The tests now measure what the blitters actually write and compare it against the width the production code chooses. The integration test also turned out to be reading a partial frame — about 74 rows of a 512-row picture — and now snapshots at paint time instead.

One finding I checked and rejected: the final blend looked like a divergence from upstream, but the standalone slang version I ported does use that select form. It is the older `glsl-shaders` file that accumulates — the same file that declares `f4` and never assigns it, which is worth knowing if anyone goes looking at the reference.

### Performance

Reported as noticeably worse on a laptop. Two things were wrong.

**It drew four times more pixels than it showed.** The mode asked for a fixed 1792x1200 drawing buffer whatever the window was doing. It now asks for the size it will actually be seen at, up to twice the usual canvas. Measured in the real app on the reporter's machine: picture displayed at 251.7 CSS pixels with `devicePixelRatio` 2.5, so 629 device pixels; before, `1792x1200` regardless; after, `896x600` — four times fewer fragments, and still twice supersampled relative to what the display can show.

**It ran the whole algorithm on flat colour.** Every rule is gated on `irlv0` — the centre differing from its right and lower neighbours — so where that is zero the answer is exactly the centre pixel. Measured over the test screens, that holds for 86% to 96% of logical pixels. Taking the exit early costs five texture fetches rather than twenty-one.

That shortcut is a branch, so the question is whether a GPU can skip anything. `irlv0` does not depend on the sub-pixel position, so it is constant across every fragment of a logical pixel — and a logical pixel is 13 to 58 output fragments on these screens, with flat regions spanning many adjacent ones. The branch is coherent far coarser than any SIMD group, so this is not a per-fragment gamble.

What keeps the shortcut honest is that the shader tests assert what the mode is *for* — flat areas untouched, hard edges hard, diagonals smoothed — so an early exit taken where the algorithm would have blended shows up as a behavioural failure. Forcing the exit to be taken always fails four of them.

Timings were taken during development with a throwaway harness rather than anything kept in tree, because headless Chrome fell back to SwiftShader on both machines tried and so only ever measured software rasterisation. Two conclusions survive that, because they are ratios: xBR is fragment-bound (1.78x the fragments cost 1.90x the time, so resolution is the right lever), and PAL is the more expensive effect at any given size by about 3.3x to 3.8x, measured independently on two machines.

### Open questions and known limits

**Is it fast enough?** Unresolved. The reporter saw it as noticeably worse before the sizing fix; whether the fix is sufficient has not been confirmed on real hardware, and cannot be from the benchmark (see above). The honest measurement is the virtual MHz readout in the app with the mode on versus off.

**At a large window this may still be slow.** The cap is `maxCanvasScale: 2`. On a high-DPI display a fuller window wants more than 2x, hits the cap, and gets no benefit from the sizing change at all — that is exactly the reporter's machine at anything but a small window. If that case matters the lever is lowering the cap: `1.5` gives 1344x900, about 1.8x fewer fragments, still comfortably above the logical grid for every BBC mode. Untested, and worth picking from a measurement rather than a guess.

**Band seams.** The shader reads each row's descriptor and samples its vertical neighbours at that row's stride, so at a mode change it reads the neighbouring mode's texels at the wrong one. Expect a row or two of artefact at a seam. Nothing on screen has looked wrong, but it is untested.

**A blend on the last row of a drawn quad.** The final row of fragments picks up a 50% blend where the picture either side of it is uniform — one pixel, where an edge meets the quad's bottom edge. It survives more sampling context and survives moving the quad edge off a logical pixel boundary, so it is neither missing context nor grid misalignment. The shader tests draw a one-pixel margin and crop it, so this falls outside what they assert; whether it is visible at the bottom of a real screen is unchecked. The old pixel-comparison would never have surfaced it — one pixel sat well inside its tolerance.

**A row records only the last mode on it.** Enough for a raster split, which is what people write; not enough for a mid-line one.

**Chunky text gets rounded.** MODE 2 and MODE 5 characters are eight logical pixels wide and xBR rounds them noticeably. Taste, not a defect.

### Test plan

- [x] `npm run lint` and `npm run ci-checks` clean
- [x] Unit, integration and shader tests all pass
- [x] New `tests/integration/pixel-grid.js` renders single-pixel stripes on a real machine in MODE 0/1/2/4/5, MODE 7 and on the Atom, and checks every texel inside one logical pixel holds the same colour, over all 512 rows
- [x] The unit tests measure the widths the blitters actually write, in all eight 6847 graphics modes plus text and all four ULA modes at both character widths
- [x] Mutation-checked: doubling the BBC width fails all five mode cases, `pixelsPerBit / bpp` fails exactly the four 2bpp Atom modes, breaking `texelsPerPixel` fails the `blitFb` test, ignoring the line grid in the shader fails exactly the three grid tests, and forcing the shader's early return fails exactly the four smoothing tests
- [x] `npm run test:shader` runs the GLSL in headless Chrome and asserts on its output; `npm test` now includes it
- [x] Existing `teletext` and `atom-hardware` integration tests pass, so the video changes are behaviour-neutral
- [x] Drawing buffer sizes above measured from the running app over CDP, per mode and window size

The other integration failures on this machine (`dormann`, `dp111timing`, `ensure-submodules`) are missing git submodules and predate this branch.

Attribution: xBR-lv2 is Copyright (C) 2011-2022 Hyllian, MIT licensed; the notice is kept in the GLSL.


src/canvas.js
```diff
@@ -179,7 +200,16 @@ export class GlCanvas {
...
 
-        this.filter.setUniforms({ width, height, frameCount });
+        this.filter.setUniforms({
+            width,
+            height,
+            frameCount: frame.frameCount,
+            lineGrid: frame.lineGrid,
+            // How much of the framebuffer each output pixel covers, which sets
+            // how wide an edge-smoothing ramp should be. Read from `extent`:
+            // minx and maxx above have been scaled into texture coordinates.
```
mattgodbolt: This comment describes the opposite of what the code does, and the code is the one that is right.

`extent` is built on line 178 from the *parameters*, before the `minx /= width` scaling on lines 187-190 — and that scaling only happens inside the `if`, on the local variables, never on `extent`'s own properties. So `extent.maxx - extent.minx` is a count of **texels** (896 for the usual visible area), and dividing by `drawingBufferWidth` gives texels per output pixel, which is exactly what the uniform is named for and what `xbr.frag.glsl` needs when it computes `aaFactor = 2.0 * uTexelsPerOutputPixel / pixelSize.x`.

If the comment were true the value would be roughly 1/1024 of the intended one, and the antialiasing ramp would collapse to nothing. The risk is that someone reads the comment, believes the value is wrong, and "fixes" working code.

Suggest something like: `extent` holds texel counts; the scaling to texture coordinates below applies only to the local copies used for the UV buffer.

src/main.js
```diff
@@ -386,11 +387,23 @@ if (keyMappingWarnings.length) {
 }
 
 function createCanvasForFilter(filterClass) {
+    const config = filterClass.getDisplayConfig();
```
mattgodbolt: `config` here shadows the module-level `const config = new Config(...)` from line 232, which is the emulator's live configuration object and is referenced throughout this file.

It works, and the previous code got away with the same name because it was scoped to the `if` block that used it. Now it covers the whole function, so anyone adding a line to `createCanvasForFilter` that wants the real `config` gets the display descriptor instead, with no error. `displayConfig` is the name already used for this thing in the resize handler further down, so it would read consistently too.

src/video-filters/shaders/xbr.frag.glsl
```diff
@@ -0,0 +1,268 @@
...
+    // sixteen texture fetches and the whole of the algorithm.
+    //
+    // Correctness rests on this being the same irlv0 the full path computes,
+    // which tools/verify-xbr-shader.js proves by comparing against the
+    // JavaScript reference, where no such shortcut exists.
+    vec4 eP4 = vec4(pack(E));
+    vec4 fP = vec4(pack(F), pack(B), pack(D), pack(H));
+    vec4 hP = fP.wxyz;
+    vec4 irlv0 = vec4(notEqual(eP4, fP)) * vec4(notEqual(eP4, hP));
+    if (irlv0 == vec4(0.0)) {
```
mattgodbolt: Worth recording why this is safe rather than leaving it implicit, because the whole performance win rests on it.

Every `texture2D` after this point is inside non-uniform control flow, which GLSL ES 1.0 leaves undefined for sampling that needs implicit derivatives (section 8.7). It is fine here for a reason that is not local to this file: `XbrFilter.getDisplayConfig` sets `nearestSampling`, so `canvas.js` gives the framebuffer texture `NEAREST` for both `MIN_FILTER` and `MAG_FILTER`, and there are no mipmaps — so no LOD is ever selected and there is nothing for the derivatives to be undefined about.

That makes the early return depend on a decision made in two other files. If anyone ever moves this filter back to `LINEAR`, or a future filter samples with mipmaps, the shortcut becomes undefined behaviour rather than merely slower. A sentence here naming that dependency would stop it being rediscovered the hard way.

The reasoning about branch coherence in the PR description is separate and holds independently.

mattgodbolt (commented):
Reviewed the whole diff. This is good work and I could not find a correctness bug in it. Three inline comments, none of which block; the summary below is what I checked and what I could not check.

### What I verified rather than assumed

- **The early-out is exactly consistent with the full path.** `irlv0` is computed once and reused after the branch, so the shortcut cannot disagree with the algorithm it is short-cutting. The `pack`/rotation bookkeeping also checks out: `fP`, `hP = fP.wxyz`, `dP = bP.yzwx` and `gP = cP.zwxy` each match the `lane(...)` ordering they stand in for.
- **`aaFactor` is dimensionally right.** `uTexelsPerOutputPixel / pixelSize.x` is output pixels per *logical* pixel, which is the unit the ramp wants. This is what convinced me the `canvas.js` comment is wrong rather than the code.
- **`updateLineGridUla` is called everywhere the mode can change.** `_writeControl` and `restoreState` are the only writers of `ulaMode`/`teletextMode`; `Ula.reset` deliberately does not touch `ulactrl`, so there is no stale-descriptor path there.
- **The grid is cleared exactly where the pixels are.** The interlaced branch of `clearPaintBuffer` clears alternate rows of both, and the doubled-row writes land on even rows, so the two stay in step.
- **`pack()` cannot collide on this hardware.** Two colours collide only if their green channels differ by less than 1/256; NuLA's finest step is 1/255, so the margin holds, if only just.

### Things I checked and decided were not worth changing

- `lineGrid[row + 1]` can write row 625, one past the picture, in both `video.js` and `6847.js`. The array is 1024 entries and `findBands` is never asked for rows past `maxy`, so nothing reads it.
- The AA ramp uses the horizontal texel ratio for vertical edges too. The output is close enough to square per logical pixel (896/896 against 600/625) that a single scalar is right.

### Not verified

I have no GPU here either, so I have taken the benchmark numbers and the shader-versus-reference agreement on trust. The 86-96% flat-pixel figure and the branch coherence argument are convincing on their face, and `tools/verify-xbr-shader.js` existing at all is the right answer to "Node has no WebGL".

The one thing I would want before merging is Matt's own measurement on the laptop that reported it slow, since SwiftShader numbers cannot settle that.

### Note on merge state

This has already picked up #764, and `shader-program.js` is a good generalisation of it — one copy of the compile/link/cleanup dance instead of three, with the failure paths preserved. Worth re-running `tests/unit/test-canvas.js` after any further rebase; it now covers the compile-failure and link-failure paths for all three filters, and those assertions depend on the error message wording that moved into `compileProgram`.

mattgodbolt:
All four addressed in ddb981c.

**The derivative dependency** — this is the one I'm glad was raised. The early return does put every `texture2D` after it inside non-uniform control flow, and GLSL ES 1.0 §8.7 leaves that undefined for lookups needing implicit derivatives. It's defined here only because there's no LOD to select: `XbrFilter` asks for `nearestSampling`, so `canvas.js` gives the texture `NEAREST` for both filters and there are no mipmaps. A shortcut in one file resting on decisions in two others is exactly the kind of thing that gets rediscovered painfully, so it's now written down at both ends — beside the return, and beside the `nearestSampling: true` that licenses it.

**`irlv0 == vec4(0.0)`** — the comparison was already correct: GLSL ES 1.0 §5.9 has `==` on vectors yield a *scalar* bool (component-wise comparison is `equal()`, which is what returns a `bvec`). That's why it has been compiling and passing the reference check throughout. But it was misread once, which is reason enough — it's `all(equal(...))` now, which can't be.

**`config` shadowing** — a real footgun, and worse than the code it replaced because it now covers the whole function rather than one `if` block. Renamed to `displayConfig`, matching the resize handler.

**The `texelsPerOutputPixel` comment** — fair. It explained the choice in terms of what had happened to the local `minx`/`maxx`, which is only true inside the branch above it. It now says what `extent` holds instead of what happened elsewhere.

782 unit tests, the integration suites and the shader-vs-reference check all still pass.

src/video-filters/shader-program.js
```diff
@@ -0,0 +1,48 @@
...
+    const program = gl.createProgram();
+    gl.attachShader(program, vertexShader);
+    gl.attachShader(program, fragmentShader);
+    gl.linkProgram(program);
+
+    if (!gl.getProgramParameter(program, gl.LINK_STATUS)) {
+        const info = gl.getProgramInfoLog(program);
+        gl.deleteProgram(program);
+        throw new Error(`Failed to link ${name} shader program: ${info}`);
+    }
```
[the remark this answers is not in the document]
mattgodbolt: Fixed. `compileProgram` now wraps compilation and linking in `try`/`finally`, so both shaders are released whichever way it exits — including the case you did not mention, where the *fragment* shader fails to compile and the vertex shader was left stranded. It matters exactly for the reason you give: `bestCanvas` retries with another filter when one throws, so a leak here accumulates.

src/main.js
```diff
@@ -386,11 +386,23 @@ if (keyMappingWarnings.length) {
...
-        showError(`enabling ${config.name} mode`, `${config.name} requires WebGL. Using standard display instead.`);
+    // Test which filter was actually built, not merely whether we got WebGL: a
+    // filter can decline a context that works perfectly well for other modes,
+    // in which case bestCanvas quietly gives us an unfiltered GL canvas.
+    if (filterClass.requiresGl() && !(newCanvas.filter instanceof filterClass)) {
+        showError(
+            `enabling ${config.name} mode`,
+            `${config.name} is not available on this device. Using standard display instead.`,
+        );
     }
```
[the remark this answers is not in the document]
mattgodbolt: Fixed, though indirectly. Every mode now declares the same 896x600 `canvasWidth`/`canvasHeight`, and xBR grows beyond that only as far as the display can actually show, via `maxCanvasScale`. So a filter that fails to build no longer leaves an oversized backing store behind — there is no longer an oversized one to leave. That change came out of a performance investigation rather than this comment, but it closes it.

src/main.js
```diff
@@ -404,7 +416,7 @@ function swapCanvas(newFilterClass) {
         frames++;
         if (frames < frameSkip) return;
         frames = 0;
-        newCanvas.paint(minx, miny, maxx, maxy, this.frameCount);
+        newCanvas.paint(minx, miny, maxx, maxy, { frameCount: this.frameCount, lineGrid: this.lineGrid });
     };
```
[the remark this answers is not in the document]
mattgodbolt: Fixed, in a separate PR so it could land on its own: #764. It was pre-existing — switching between RGB Monitor and PAL TV already stranded a 1024x1024 RGBA texture, two buffers, a program and its shaders per switch — so it did not belong in this branch. `GlCanvas.dispose()` releases what the canvas owns and asks the filter to do the same, and `swapCanvas` calls it after the replacement is built. `XbrFilter.dispose()` also releases the extra line-grid texture, and `tests/unit/test-canvas.js` runs the disposal check over all three filters.

src/canvas.js
mattgodbolt: Reworded, near enough to your suggestion. It now says `extent` holds texel counts and that the scaling into texture coordinates applies only to the local copies going into the UV buffer — describing what the value *is*, rather than what happened to two other variables. Your point about the conditional is the sharper one: the scaling only happens inside the `if`, so the old wording was not merely backwards but intermittently true, which is worse.

src/main.js
mattgodbolt: Renamed to `displayConfig`, matching the resize handler. Agreed it is worse than what it replaced: the old one was scoped to the `if` block that used it, and mine covered the whole function, so anyone adding a line wanting the real `config` would have silently got the display descriptor. There is now a comment saying which `config` is meant.

src/video-filters/shaders/xbr.frag.glsl
mattgodbolt: Written down, at both ends. The note beside the early return names GLSL ES 1.0 section 8.7 and spells out why it is defined here — `XbrFilter` asks for `nearestSampling`, so `canvas.js` gives the texture NEAREST for both filters and it has no mipmaps, hence no LOD to select and nothing for the derivatives to be undefined about — and says what breaks the reasoning. There is a matching note beside `nearestSampling: true` pointing back, since that is where someone would change it from.

This was the most useful comment of the four: a shortcut in one file resting on decisions in two others, with nothing recording the dependency. Agreed the branch-coherence argument is separate and stands on its own.

mattgodbolt:
Bug: when switching between display types it used to be the case that the image updated (behind the settings window). It now goes black.

mattgodbolt:
Split into a stack, without rewriting any history on this branch: the parts that stand on their own are now #779 (one place to compile a shader program), #780 (the `bestCanvas` fallback fix) and #781 (a name in the render loop). This branch merges them in and sits on top, so the diff here is xBR only, and reverting it later leaves those in place.

Left here because they only do anything once a mode scales its drawing buffer: `GlCanvas` tracking the viewport, and `Canvas.paint` reading the element size each time.

The merge also picked up `testTimeout` from main, which this branch had as `testTimout` and so was never applying.

PR #776, opened 2026-08-08
Fix the display filter fallback, and compile shaders in one place

mattgodbolt:
Three independent changes to the display filter plumbing.

### A filter that will not build no longer takes the emulator with it

`bestCanvas` catches a failure from `new GlCanvas(...)` and falls through to `new Canvas(canvas)`. If the failure came from WebGL being unavailable, that is right. If the context was created and the *filter* failed to compile or link, it is not: the canvas element has already handed out a WebGL context and can never hand out a 2D one, so the fallback throws out of `bestCanvas` and the emulator does not start at all.

It now retries the passthrough filter on the context it already has, and only reaches the 2D canvas when there was no WebGL to begin with. `tests/unit/test-canvas.js` covers it by failing the PAL fragment shader and nothing else; with the fallback removed that test throws, which is the bug.

Canvases gained a `filterClass` getter reporting what they actually built, replacing `isWebGl()`. That is also a truer test for whether to tell the user their mode is unavailable, since a filter can decline a context other modes are perfectly happy with. `requiresGl()` then had no callers left and is gone.

### One place to compile a shader program

Both filters kept their own copy of the compile, link and cleanup dance, and any future one would need a third. It moves to `src/video-filters/shader-program.js`, which takes the filter name so a failure still says which shader it was. No behaviour change beyond the wording of two error messages and the loss of a `console.log` on PAL init.

### A name in the render loop

The render loop reused one variable for the scanline and then for the byte offset derived from it.

### Test plan

- [x] `npm run lint` clean
- [x] 750 unit tests pass, including the new fallback case
- [x] Integration suite passes (15 files, 74 tests)
- [x] The new test is mutation-checked: removing the fallback makes it fail

### DIFF
--- a/src/canvas.js
+++ b/src/canvas.js
@@ -13,8 +13,9 @@ export function getFilterForMode(mode) {
 }
 
 export class Canvas {
-    isWebGl() {
-        return false;
+    /** The 2D canvas draws the framebuffer as-is, which is what this filter is. */
+    get filterClass() {
+        return PassthroughFilter;
     }
 
     constructor(canvas) {
@@ -46,8 +47,9 @@ export class Canvas {
 const width = 1024;
 const height = 1024;
 export class GlCanvas {
-    isWebGl() {
-        return true;
+    /** The filter actually built, which may not be the one that was asked for. */
+    get filterClass() {
+        return this.filter.constructor;
     }
 
     constructor(canvas, filterClass) {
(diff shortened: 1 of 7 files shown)


mattgodbolt:
Split into one PR per change: the shader compile module, the fallback fix, and the render loop naming. Closing in favour of those.

PR #781, opened 2026-08-08
refactor: name the scanline and the framebuffer offset separately

mattgodbolt:
The render loop reused one variable for the scanline and then for the byte offset derived from it, in the hottest loop in the emulator.

Independent of the rest of the stack; it sits here only to keep one chain.


mattgodbolt:
Good catch, thank you. `row` does read as a character row in this file, and `scanline` would have collided with `scanlineCounter`, which is the CRTC's counter within a character row. Renamed to `bitmapRow`, which pairs with `bitmapX` and `bitmapY`.

PR #222, opened 2019-08-23
Add a text field which converts paste events to BBC key sequences.

scarybeasts:
Fixes #164 

Supports paste via Ctrl-V, middle-mouse (on Linux at least) and context menu -> paste.

PR #842, opened 2026-08-13
fix: stop the unstyled flash on load

mattgodbolt:
The whole UI lives in `index.html` and is laid out entirely by CSS, so any paint before the stylesheet applies shows every modal and debug panel as a wall of text. Always visible on the dev server (Vite injects the CSS from JS); in production it needs a slow connection or a browser with a FOUC timeout, since the stylesheet is render-blocking there.

An inline `<style>` in the head hides the body and paints `#222` (bootswatch darkly's `--bs-body-bg`) from the first frame; `jsbeeb.css` reveals it. If the stylesheet ever fails outright you now get a dark blank page rather than the wall of text.

Separately: everything was deployed with `max-age=30`, including the content-hashed `assets/`, so every visit refetched the CSS and JS. Those now get an immutable year-long cache, and upload first so `index.html` can't reference a file that isn't in the bucket yet.

Checked on a local build that a normal load is unchanged, and that a copy with the stylesheet link stripped out stays dark and empty.

### DIFF
--- a/.github/workflows/test-and-deploy.yml
+++ b/.github/workflows/test-and-deploy.yml
@@ -84,6 +84,11 @@ jobs:
           AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
           AWS_REGION: us-east-1
         run: |
-          aws s3 sync dist s3://bbc.xania.org \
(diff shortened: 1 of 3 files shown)


mattgodbolt:
Split into #844 (fix) and #845 (ci) so the squash-merge keeps the conventional-commit types distinct.

Issue #865, opened 2026-08-15
Audio: buffer management drops audio on ordinary frame jitter, allocates on the audio thread, and cannot underrun without a click


mattgodbolt:
Part of #861. Five related defects in the queue and its consumption, all of which get smaller or vanish if the queue becomes a flat ring buffer.

## The drop threshold is below one frame of production

`targetLatencyMs = 20` ("one frame") and `maxLatency = targetLatencyMs * 2 = 40 ms`. But the producer is not continuous: `main.js:2340` calls `soundChip.catchUp()` once per animation frame, so at 60 Hz the chip emits ~16 chunks of 512 samples in one burst every 16.7 ms and nothing in between.

That makes 20 ms of target *one burst plus 3.3 ms*, and the 40 ms drop threshold *one burst plus one target*. Simulated over 30 s at 48 kHz, reproducing `onBuffer`, `cleanQueue`, `nextSample`, `process` and the 1 ms `Date.now()` quantisation:

```
60Hz clean                    underruns 0       silent quanta 0     dropped 0     min age 8.8ms
60Hz, 1 missed vsync in 60    underruns 0       silent quanta 0     dropped 37    min age 3.8ms
60Hz, 1 missed vsync in 10    underruns 0       silent quanta 0     dropped 44    min age 3.5ms
30Hz clean                    underruns 0       silent quanta 0     dropped 0     min age 0.0ms
30Hz, 1 missed vsync in 30    underruns 11190   silent quanta 286   dropped 800
```

**A single missed vsync** in an otherwise clean 60 Hz run is enough to start discarding: the doubled frame delivers 33 ms of audio at once and the head of the previous burst ages past 40 ms before it can be consumed. Each drop splices out 1–2 ms at a hard edge. At 30 fps the same jitter gives 286 fully silent quanta — 0.76 s of silence — and 800 dropped chunks in 30 s.

The two constants are coupled to the producer's burst size and should be derived from it rather than from "one frame". Target wants to be at least one burst plus jitter margin; the drop threshold is an emergency valve and belongs several bursts above target, not at 2×. Alternatively make the producer non-bursty by calling `catchUp()` several times per frame, which lowers the required target directly.

`dropped` is posted in the stats message but never charted (`audio-handler.js:101-102` adds only `queueSize` and `queueAge`), which is why none of this has been visible.

## The audio thread allocates ~4 MB/s

[...]

Issue #862, opened 2026-08-15
Audio: there is effectively no anti-alias filter, worst in-band spurs around -25 dBc


mattgodbolt:
Part of #861.

The chain decimates 500 kHz to 48 kHz, a ratio of 10.4, with exactly two filters:

- a one-pole RC at `sampleRate / 2` = 24 kHz, run at the input rate (`audio-renderer.js:3-4`)
- linear interpolation between adjacent **input** samples, whose kernel is two input samples wide and whose first null is therefore at 500 kHz. At 46 kHz its response is −0.23 dB.

So the only thing between 500 kHz of square-wave harmonics and the 24 kHz fold point is a single pole sitting *on* the fold point: −3 dB at Nyquist, −7 dB at 48 kHz (which folds to DC), −12 dB at 100 kHz (folds to 4 kHz). A decimate-by-10 wants 40–60 dB there.

## Measured

One tone channel at full volume, 48 kHz device rate. Columns are total non-harmonic power relative to harmonic power, and the worst single spur, in dBc:

```
reg   f0(Hz)   worklet          no RC at all     box average      reference FIR
1000     125    -25.9/ -38.6    -23.1/ -37.8    -67.5/ -63.7    -67.2/ -63.7
 284     440    -28.6/ -38.8    -21.9/ -35.0    -31.2/ -39.2    -43.0/ -60.8
 100    1250    -24.5/ -28.6    -17.6/ -24.6    -26.9/ -30.5    -47.4/ -53.5
  50    2500    -21.3/ -24.5    -14.5/ -20.2    -23.8/ -25.8    -47.6/ -52.3
  20    6250    -17.0/  -7.4    -10.3/  -2.8    -20.6/ -20.9    -30.4/ -37.7
```

**Deleting the RC entirely costs about 6 dB**, which is what a single pole on the fold point is worth. It is decorative.

The residue is not small, and it is *inharmonic*. For a 1250 Hz tone the largest spurs are:

```
21757 Hz  -28.6 dBc      1761 Hz  -37.2 dBc
19257 Hz  -30.7 dBc       738 Hz  -38.4 dBc
16758 Hz  -31.1 dBc      3237 Hz  -38.9 dBc
```

Odd harmonics 33/35/37/39 folding around 48 kHz. Because 48 kHz is not harmonically related to the note, they land between the real partials and move in the *opposite* direction when the note changes.

[...]

PR #881, opened 2026-08-25
fix: stop the audio rate control warbling at the display frame rate

### LINKED ISSUE #864: Audio: the rate control loop modulates pitch with the producer's frame-rate sawtooth
Part of #861.

## The loop is stable; the measurement is noisy

Writing `L` for queue age, production is real-time and consumption is `effectiveSampleRate`, so with `adjustment = 100 · (L − Lt)` Hz per ms the response is `dL/dt = −0.2 (L − Lt)`: first order, 5 s time constant, against a measurement delay of ~20 ms. Massively over-damped. There is no stability problem.

The problem is that the proportional term acts on an *instantaneous, unfiltered* measurement, and that measurement carries the producer's burst sawtooth at full amplitude. `main.js:2340` calls `soundChip.catchUp()` once per animation frame, so at 60 Hz the chip emits ~16 chunks of 512 samples in one burst every 16.7 ms and nothing in between.

Simulated steady state, 60 Hz producer, 48 kHz device:

```
t(ms)     queueAge(ms)   playback rate (cents from nominal)
1013.33       30            +3.31
1016.00       15            +4.14   <- burst arrives
1018.67       19            -0.57
1029.33       29            +3.44
1032.00       15            +3.91   <- burst arrives
```

Peak-to-peak over 20 s: **6.7 cents at 60 fps, 3.4 at 144 fps, 12.2 at 30 fps**, as a sawtooth at the frame rate. That is frequency modulation of everything at the refresh rate, peak deviation ±0.15 %, sidebands at ±60 Hz around every partial: roughly −38 dBc for a 1 kHz tone, −26 dBc at 4 kHz, −20 dBc at 8 kHz.

**Predicted effect:** roughness on high tones and a slight warble on sustained notes, correlated with the display refresh rate. Unconfirmed by ear.

## `_queueAge()` measures the wrong thing

It compares `Date.now()` in the worklet against a timestamp taken when the producer posted the buffer. Both ends read the same system clock so they *are* comparable — that part is fine. The problems are:

- the timestamp is the **postMessage time, not the generation time**, which is exactly what turns a burst into a sawtooth
- `Date.now()` is **not monotonic**: a −50 ms clock step parks the resampler at its −1 % clamp, −17 cents, until the queue drains
- 1 ms quantisation against a 20 ms target is 5 % of the control variable

Tracking **sample occupancy** — how many samples are queued, which the worklet already knows — removes all three. It needs no clock at all.

## Fix

Low-pass the measurement before the controller. The plant has a 5 s time constant, so an exponential average with a 0.3–1 s time constant costs nothing in responsiveness and removes essentially all of the frame-rate content. Better, use PI on smoothed occupancy: P to reject transients, I to absorb the constant device-versus-emulator clock offset, which is a few ppm and does not need a 100 Hz/ms proportional term to track.

Trade-off: a slow loop recovers from genuine excursions over seconds rather than being caught by the drop path, which is what you want given #865.

Related: `maxAdjust = inputSampleRate * 0.01` (`audio-renderer.js:87`) is unreachable dead code. It clamps at |outByMs| > 50 ms, but `cleanQueue` drops anything over 40 ms and `_queueAge()` returns ≥ 0, so the achievable range is [−20, +20].

mattgodbolt:
Fixes #864. **Numerically verified but needs human listening before merge**; instructions below.

## What this does

The resampler's rate controller acted on an instantaneous queue-age measurement that carries the producer's once-per-frame burst sawtooth at full amplitude, so the playback rate was frequency modulated at the display refresh rate: ~6.7 cents peak to peak at 60 fps, worse at lower frame rates (the issue's analysis). The measurement also stamped postMessage time rather than generation time, and used the non-monotonic `Date.now()` (a -50 ms clock step could park the resampler at its -1% clamp).

Now the controller runs on **queue occupancy in samples** (which the worklet already tracks, and which needs no clock at all), smoothed with a 0.5 s time constant. The loop keeps its existing 5 s response; the smoothing removes essentially all frame-rate content. Simulated with the real burst pattern: 60 Hz wobble drops from ~6.7 cents to **0.01 cents**.

One deliberate deviation from the issue: **proportional only, not PI.** Occupancy already integrates any rate offset, so adding an integral term makes the loop second order; at plausible gains the simulation overshot into near-underrun and took ~40 s to settle. The offset a PI would remove costs under a millisecond of static latency error at realistic clock skew, which is not worth that behaviour.

A new test harness (`tests/unit/test-audio-renderer.js`) runs the worklet processor in Node against the real 60 Hz burst pattern and pins: no frame-rate modulation, correct control direction, and convergence with zero drops and zero underruns.

Not touched (that's #865): the age-based drop policy in `cleanQueue`, the buffer allocation on the audio thread, and the underrun click.

## What to listen for

**Should sound better (the point of the change):**
- **Sustained tones and high notes.** Long held notes (title-screen music, the BASIC `SOUND 1,-15,200,200` sort of thing, anything with slow melodies) previously carried a subtle roughness or warble correlated with your display refresh rate. Listen at 60 Hz and, if convenient, at 144 Hz and with the window backgrounded or the machine loaded, where the old behaviour differed; the new one should not.
- **High-pitched content** (above ~4 kHz) had the strongest FM sidebands; it should sound cleaner.

**Must not regress:**
- **Latency and stability.** The loop's recovery behaviour is deliberately unchanged (5 s time constant); pause/resume the emulator, drag the window around, run something CPU-heavy alongside. No new clicks, dropouts, or drift over a long session (leave a tune looping for ten minutes).
- **Sampled speech and everything else**: this change is upstream-neutral; anything that sounds different beyond the warble is a bug.

**Against the real Master:** a long sustained tone on both, ideally the same note; the Master obviously has no frame-rate warble, and now neither should jsbeeb.

https://claude.ai/code/session_019CJ9csHch7kjzF3DKzR5DP

### DIFF
--- a/src/web/audio-renderer.js
+++ b/src/web/audio-renderer.js
@@ -16,9 +16,13 @@ const samplesFor = (ms) => (InputSampleRate * ms) / 1000;
 // an integral term makes the loop second order (overshooting into underrun
 // and settling in ~40 s at plausible gains), and the offset it would remove
 // costs under a millisecond of static latency error at realistic clock skew.
+// The authority is small: clock skew and the resampler's per-quantum rounding
+// are both under 0.04%, and a slow loop with the old 1% clamp turned queue
+// disturbances (a double burst, a dropped buffer) into pitch bends of several
+// cents held for seconds. Anything the clamp cannot absorb, the queue does.
 const OccupancySmoothingTau = 0.5;
 const ProportionalGain = 0.2;
-const MaxAdjust = InputSampleRate * 0.01;
+const MaxAdjust = InputSampleRate * 0.001;
 
 class SoundChipProcessor extends AudioWorkletProcessor {
     constructor(...args) {
--- a/src/web/audio-renderer.js
+++ b/src/web/audio-renderer.js
@@ -8,6 +8,18 @@ const MaxQueuedMs = 250;
 
 const samplesFor = (ms) => (InputSampleRate * ms) / 1000;
 
+// Rate control: proportional control on queue occupancy, smoothed to reject
+// the producer's once-per-animation-frame burst pattern (issue #864). The
+// loop's time constant is seconds (gain 0.2/s, the old controller's pole), so
+// half a second of smoothing costs nothing in responsiveness. Proportional
+// only, deliberately: occupancy already integrates any rate offset, so adding
+// an integral term makes the loop second order (overshooting into underrun
+// and settling in ~40 s at plausible gains), and the offset it would remove
+// costs under a millisecond of static latency error at realistic clock skew.
+const OccupancySmoothingTau = 0.5;
+const ProportionalGain = 0.2;
+const MaxAdjust = InputSampleRate * 0.01;
+
 class SoundChipProcessor extends AudioWorkletProcessor {
     constructor(...args) {
         super(...args);
@@ -21,6 +33,7 @@ class SoundChipProcessor extends AudioWorkletProcessor {
         this.underruns = 0;
         this.targetLatencyMs = 1000 * (1 / 50); // One frame
         this.startQueueSizeSamples = samplesFor(this.targetLatencyMs);
+        this.smoothedOccupancyError = 0;
         this.running = false;
         this.maxQueueSizeSamples = samplesFor(MaxQueuedMs);
         this.port.onmessage = (event) => {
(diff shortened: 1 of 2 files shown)


mattgodbolt:
Applied the suppressed review observation about the harness clock: `simulate()` now drives `Date.now()` from simulated time, so the age-based cleanup sees the same timeline as the controller and a wall-clock stall (GC pause past the 40 ms cap) can no longer drop buffers mid-test. One knock-on correction: with a faithful clock, the convergence test's old 2x-target prefill would have been shed by the age cap rather than the controller (post-burst backlog over 20000 samples exceeds 40 ms of age), so it now starts one target's worth over target, the largest excess the drop policy leaves to the controller. That the age cap beheads anything larger is #865's territory, not this PR's.

mattgodbolt:
Listening found a regression: a held `SOUND 1,-15,200,200` warbled, where live is rock solid. Cause: the old controller reacted within one 2.7 ms quantum, so whatever it did to pitch lasted milliseconds; this loop is deliberately slow (that is the whole fix), so the same queue disturbances (the initial fill converging from below target, a double burst after a dropped frame, the age cap shedding buffers) became pitch bends of several cents held for seconds, up to the 1% clamp.

Fix pushed in 2221ed4: clamp the controller's authority at 0.1% (under two cents) instead of 1%. Clock skew and the resampler's per-quantum rounding are both under 0.04%, which is all rate control legitimately has to track; anything bigger is a queue-level problem and the queue absorbs it, as on main. A test now pins the limit in both directions.

Listening criterion is unchanged and now sharper: a held tone must be at least as steady as on live. If it is not clearly so, this PR should be closed rather than merged, since #864's audible premise ("unconfirmed by ear" in the issue) has not survived contact with ears on the old controller either.

mattgodbolt:
Second listening pass: a held tone started steady, then kinked down about 3 s in, at the moment main produces its first underrun click (main's pitch stays constant through the click).

Cause, in 426eb66: the resampler rounded its per-quantum consumption to whole input samples, so its actual pitch was quantised in steps of one part in ~1300, about 1.3 cents. The old controller's frame-rate jitter dithered that quantisation into inaudibility; a steady rate sits on one side of a boundary until a queue perturbation (the underrun) nudges it across, giving a clean 1.3 cent step. Fix: carry the fractional read position across quanta so consumption averages exactly the ratio, with no boundary to cross (this also removes the per-quantum phase jump, one of #862's spurs, and the per-quantum allocation on the audio thread). With no rounding deficit left to track, the controller's authority is now 0.05%, under a cent, so even a full-clamp excursion is below the pitch JND. Tests pin exact consumption and the authority limit.

PR #896, opened 2026-08-26
fix: initialise Google Drive on the first click that needs it, not at boot

### LINKED ISSUE #888: Google Drive client initialises on the main thread at page load, stalling the emulator for ~100ms during boot
At page load, `src/main.js` eagerly runs `googleDrive.initialise()` (to decide whether to show the `.if-drive-available` elements and attempt a silent auth). That loads `apis.google.com/js/api.js` and `accounts.google.com/gsi/client`, creates the token client, loads the `client` module and runs `gapi.client.init` with the Drive discovery document, all on the main thread while the emulator is already booting.

Seen with the `?audioDebug` tick log from #886/#887, on every fresh load: the second in which "Client loaded; initialising GAPI" / "GAPI initialised" print is the one with the largest main-thread hold-off of the whole session, typically `idle max 100-130ms`, and it comes with a burst of audio underruns and an audible stutter during the boot beep or the first thing the disc plays. Nothing else in the startup log is in that range once the JIT has warmed.

Most sessions never touch Google Drive, so this cost is paid for nothing. Options, cheapest first:

- Defer the whole initialisation until the page is idle (`requestIdleCallback`, or simply a few seconds after boot), keeping the `.if-drive-available` elements hidden until then. The Drive menu items would appear a moment later than now.
- Or initialise lazily on first use (the load path in `loadGoogleDrive` already calls `initialise()` itself), and decide `.if-drive-available` from a much cheaper signal, or just show the items and let the click do the work.
- Either way, `gapi.client.init` with a discovery document is the expensive step; it could be done only when a Drive action is actually requested.

Related: #885 (the emulator loop no longer depends on the display, so startup hold-offs like this are now the visible remainder).

mattgodbolt:
Fixes #888. Loading Google's client scripts and running `gapi.client.init` with the Drive discovery document held the main thread for ~100ms during every boot (the `idle max 100-130ms` line that came with a boot-time stutter), for a feature most sessions never use.

## What changed

- Nothing Google-related happens at boot. The "From Google Drive" menu item is always shown.
- Clicking it runs `googleDrive.initialise()` (memoised, so repeated clicks and the `gd:` disc-URL path share one load, and a failure clears the memo so a later click can retry), then the existing sign-in flow (silent first, then prompting) and the modal. If the scripts are blocked (an extension, offline) the click says so in a toast.
- The Drive save form is only reachable from that modal, so it is always initialised by the time it runs.

An earlier version of this PR delayed the initialisation by 3s; that only moved the stall, as reviewed.

## What to check

- With `?audioDebug`, the boot second should show a small `idle max` (single digits here) and no "Creating GAPI" in the console.
- "From Google Drive" opens the modal after a short wait the first time, lists files, loads one; creating a disc from the modal still works.
- A `disc=gd:...` URL still loads.

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -1725,22 +1725,16 @@ async function gdLoad(cat, layout) {
     }
 }
 
-for (const el of document.querySelectorAll(".if-drive-available")) el.style.display = "none";
-// Loading the Google client holds the main thread for ~100ms; keep that clear of boot.
-const GoogleDriveInitDelayMs = 3000;
-async function showGoogleDriveWhenAvailable() {
-    try {
-        const available = await googleDrive.initialise();
-        if (available) {
-            for (const el of document.querySelectorAll(".if-drive-available")) el.style.display = "";
-            await gdAuth(true);
-        }
-    } catch (error) {
-        console.log(`Google Drive is unavailable: ${errorText(error)}`);
-    }
-}
 const googleDriveModal = new bootstrap.Modal(googleDriveEl);
+// Loading the Google client holds the main thread for ~100ms, so it waits for
+// someone to ask for Drive.
 document.getElementById("open-drive-link").addEventListener("click", async function () {
+    try {
+        await googleDrive.initialise();
+    } catch (error) {
+        toast(`Google Drive is unavailable: ${errorText(error)}`, { title: "Google Drive" });
+        return false;
+    }
     const authed = await gdAuth(false);
     if (authed) {
         googleDriveModal.show();
@@ -2208,7 +2202,6 @@ const startPromise = (async () => {
         }
 
         go();
-        window.setTimeout(showGoogleDriveWhenAvailable, GoogleDriveInitDelayMs);
     } catch (error) {
         console.error("Error initialising emulator:", error);
         showError("initialising", error);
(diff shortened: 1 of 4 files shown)


src/main.js
```diff
@@ -1725,7 +1725,9 @@ async function gdLoad(cat, layout) {
 }
 
 for (const el of document.querySelectorAll(".if-drive-available")) el.style.display = "none";
-(async () => {
+// Loading the Google client holds the main thread for ~100ms; keep that clear of boot.
```
mattgodbolt: this doesn't fix the problem, it just moves it to a later moment

mattgodbolt (commented):
we need to discuss a better fix

src/main.js
mattgodbolt: Agreed. Proposal: no initialisation at boot at all. Show the Drive menu item always; on the open-drive-link click, await googleDrive.initialise() (memoised, so it is shared with the gd: disc-URL path, which already initialises itself), then the existing gdAuth(false), which tries silently before prompting, then the modal. If Google's scripts are blocked the click reports it through the modal's existing error text instead of the item being hidden. The stall then only ever happens on the click of someone who wants Drive, once per page load, and most sessions never pay it. Doing it before go() instead would make boot wait on two remote script loads; splitting it still evaluates Google's script mid-play for nothing. Say the word and I will rework the branch that way, replacing the delay.

mattgodbolt: That sounds much better; agreed.

Issue #892, opened 2026-08-26
Music 5000 worklet has no queue management, and its AudioContext is created whether or not one is fitted


mattgodbolt:
Split out of #865, which covered the SN76489 worklet and is otherwise resolved. The Music 5000 path (`src/music5000-worklet.js`, `src/web/audio-handler.js`) has none of the queue management the main worklet now has:

- No start threshold, occupancy check or underrun detection; the reader cannot tell whether the data under it was written this lap or is a lap stale. With a bursty producer (now 10 ms ticks) and a smooth consumer, the reader can overtake the writer and replay the previous lap.
- The ring wrap relies on the bounds check happening once per message before a whole block is written, which is only safe because 256 and the 128-frame quantum both divide 65536; changing either writes past the end and reads `undefined` into the output.
- It allocates per message.
- `audio-handler.js` creates a second `AudioContext` at 46875 Hz with the worklet connected on every session, whether or not a Music 5000 is fitted: a second device stream and a second browser resampler for everyone.

The last point is the cheap one (create the context lazily when a Music 5000 is enabled); the rest wants the same treatment the main worklet got in #886/#891, or sharing its queue code.

mattgodbolt:
Very much less worried about this - music5000 support is super niche (I don't even really know how to test it), but this is for completeness.

PR #897, opened 2026-08-26
feat: synthesise the sound chip in the audio worklet from timestamped register writes

### LINKED ISSUE #894: Audio: synthesise the sound chip in the worklet from timestamped register writes, so a stall stretches the sound instead of cutting it
Today the main thread renders the SN76489 to samples and ships ~1000 buffers a second to the worklet; whatever stalls the main thread stalls the sample supply, and the best the worklet can do is fade (#893). If the worklet ran the synthesis itself from a stream of timestamped register writes, the thing that could run out would be *future writes*, and the failure mode changes from a gap to a stretch: a held tone stays held, a tune's current note lasts a little longer, PCM speech sits on its (inaudible) carrier until the writes arrive.

Sketch, from the current design:

- The main-thread `SoundChip` stays as it is (snapshots, headless harness, MCP `read_sound_state`, tests all use it); in the browser it stops generating samples and `catchUp()` leaves the tick.
- Each write that reaches the chip (after the scheduler's 32-cycle slow-bus delay, so the effective time is the stamp) is posted as `{cycle, value}`, batched per tick. Far fewer messages and bytes than sample buffers, even for the PCM tricks.
- The worklet imports `soundchip.js` (AudioWorklet modules are ES modules) and owns a second instance, advancing its clock at 500 kHz, applying writes when its clock reaches their stamps, and resampling as now.
- The rate controller watches the lead of the newest write over the worklet's clock instead of queue occupancy, aiming for the same ~20 ms with the same 0.05% authority. When lead reaches zero the worklet simply keeps rendering the current state.
- The open design question: when late writes arrive, apply them immediately (a tempo blip, no pitch bend) or slew the worklet clock to absorb them (no blip, a tiny bend). Needs listening.
- Save-state restore and rewind must resynchronise the worklet's chip with a full-state message (`snapshotState` already produces it); `?audioDebug` reports lead instead of occupancy.

Music 5000 would want the same separately (#892).

mattgodbolt:
Implements #894. The browser's sound is now synthesised inside the audio worklet from the chip's register writes, so a stall on the main thread stretches the sound instead of cutting it.

## What changed

- **The chip emits events instead of samples when it has an event sink.** `SoundChip` takes an `onEvent` option; every state change (register write, tape tone on/off, mute/unmute, reset, restore, and the Atom's speaker bits) is reported stamped with the scheduler cycle it takes effect at, and a chip with a sink does not render. The main-thread chip keeps everything else it had (snapshots, `debugPokeAll`, the headless harness's `InstrumentedSoundChip`, the tests), so nothing outside the browser changes.
- **The handler ships events once per tick** (`flushChipEvents`, called from the tick after `execute()` and after a blur run-ahead), together with the emulator's current epoch, so the worklet always knows how far ahead the producer is even when nothing changed. That replaces ~1000 sample buffers a second with ~100 small messages.
- **The worklet owns a second `SoundChip`** (imported; the renderer is now bundled as a worker via `?worker&url` because a plain asset URL copies a worklet verbatim) and renders from the events, applying each when its clock reaches the event's cycle, at sample (2 µs) granularity, exactly as the main-thread chip used to. Resampling and the 0.05% rate control are unchanged apart from watching *lead* (how far the emulator is ahead of the worklet's clock) rather than queued samples.
- **Stalls and skips replace underruns and drops.** When the worklet's clock reaches the producer's position it holds the chip's state and its clock: a held tone keeps sounding, a note lasts longer, sampled speech sits on its inaudible carrier. When the producer is a full target ahead again the missed writes are applied in one go and the clock jumps, so the sound continues from the emulator's current state with no gap and no waveform tear. A restore or reset is a resync: the worklet jumps to the new timeline and discards anything queued from the old one, which is what rewind and save-state loading need.
- `?audioDebug` now charts lead and shows stalls (red) and skips (purple, height = ms skipped); the console line reads `audio lead min Nms, stalls N, skipped Nms`. `audioLatencyMs` now means target lead. #889's blur/focus depth change works unchanged (the run-ahead just moves the emulator's position).

Tests: the worklet is driven with the real `SoundChip` by a simulated producer and checked with Goertzel: the tone renders at the right frequency, a write lands at its cycle mid-quantum, the tone carries on through a stall, a burst catch-up skips the right amount, a producer pause that resumes at real time needs no skip, a rewind resyncs, rate control converges and stays within authority. Chip tests cover the events and that rendering from them matches the source chip; the handler test covers the flush. Verified headless in dev and in the production build (`vite preview`), including a blur/focus cycle.

## What to listen for

1. `SOUND 1,-15,200,200` with `?audioDebug`. Previously an underrun was a gap; now a stall should be inaudible on a held tone (the tone continues) and the line shows `stalls N, skipped 0ms` for brief ones. Provoke long ones with `audioLatencyMs=5`: a stall then shows as `skipped Nms`, which on a held tone is still nothing to hear.
2. Music (a demo with a tune): a long stall should now sound like the current note hanging for a moment and the tune then jumping on, never a click or a gap.
3. **Sampled speech and the PCM tricks (Speech!, Spyhunter, Repton, Chris's demos)**: these are where the timing of writes matters most. They should sound exactly as before in normal running; a stall becomes a short dropout to the carrier rather than a gap. This is the case I can least verify without ears.
4. Tape loading tones (the sine channel), and the Atom if you have one to hand: both now travel as events.
5. Save state, load state, rewind, hard and soft reset: the sound should follow the machine without a burst of stale audio.
6. Alt-tab away and back (#889 behaviour should be unchanged), pause and resume.

Closes #894. The Music 5000 (#892) still uses the sample path.

### DIFF
--- a/src/soundchip.js
+++ b/src/soundchip.js
@@ -12,6 +12,10 @@ export const SoundBufferSamples = 512;
 // mean level, which a zero-mean output would silence (see issue #863).
 const DcRestoreCornerHz = 1 / (2 * Math.PI * 10e3 * 4.7e-6);
 
+// A chip with an event sink reports progress this often, so its events can
+// stream through a long execute() rather than all arriving at its end.
+const EventProgressCycles = 4000;
+
 const volumeTable = new Float32Array(16);
 (() => {
     let f = 1.0;
@@ -249,6 +253,13 @@ export class SoundChip {
         this.activeTask = this.scheduler.newTask(() => {
             if (this.active) this.poke(this.slowDataBus);
         });
+        if (this._onEvent) {
+            this.progressTask = this.scheduler.newTask(() => {
+                this._emit({ progress: true });
+                this.progressTask.schedule(EventProgressCycles);
+            });
+            this.progressTask.schedule(EventProgressCycles);
+        }
     }
 
     render(out, offset, length) {
@@ -382,6 +393,7 @@ export class SoundChip {
         // Older snapshots predate the DC blocker
         this.dcPrevIn = state.dcPrevIn ?? 0;
         this.dcPrevOut = state.dcPrevOut ?? 0;
+        this.progressTask?.ensureScheduled(true, EventProgressCycles);
         this._emit({ state: this.snapshotState() });
     }
 
--- a/src/soundchip.js
+++ b/src/soundchip.js
@@ -35,9 +35,15 @@ export class SoundChip {
      * @param {function(Float32Array): void} onBuffer called with each full
      *     SoundBufferSamples-sized buffer of output. Receives the same buffer
      *     every call, overwritten afterwards: copy the contents if they are kept.
+     * @param {object} [options]
+     * @param {function(object): void} [options.onEvent] called with each
+     *     change to the chip's state, stamped with the cycle it takes effect at.
+     *     A chip with an event sink does not render; something else renders
+     *     from the events (see audio-renderer.js).
      */
-    constructor(onBuffer) {
+    constructor(onBuffer, { onEvent = null } = {}) {
         this._onBuffer = onBuffer;
+        this._onEvent = onEvent;
         // 4MHz input signal. Internal divide-by-8
         this.soundchipFreq = 4000000.0 / 8;
         const sampleRate = this.soundchipFreq;
@@ -91,13 +97,36 @@ export class SoundChip {
             mute: () => {
                 this.catchUp();
                 this.sineOn = false;
+                this._emit({ sine: 0 });
             },
             tone: (freq) => {
                 this.catchUp();
                 this.sineOn = true;
                 this.sineStep = (freq / sampleRate) * this.sineTable.length;
+                this._emit({ sine: freq });
             },
         };
+    }
+
+    _emit(event) {
+        if (this._onEvent) this._onEvent({ cycle: this.scheduler.epoch, ...event });
+    }
+
+    /** Applies an event from another chip's onEvent, at the cycle this chip is rendering. */
+    applyEvent(event) {
+        if (event.poke !== undefined) this.poke(event.poke);
+        else if (event.sine !== undefined) {
+            if (event.sine) this.toneGenerator.tone(event.sine);
+            else this.toneGenerator.mute();
+        } else if (event.enabled !== undefined) this.enabled = event.enabled;
+        else if (event.state !== undefined) this.restoreState(event.state);
+        else if (event.reset !== undefined) this.reset(event.reset);
+    }
+
+    /** Renders `length` samples of output from `cycle`, for a chip that is driven by events. */
+    renderAt(cycle, out, offset, length) {
+        this.scheduler.epoch = this.lastRunEpoch = cycle;
+        this.generate(out, offset, length);
     }
 
     sineChannel(channel, out, offset, length) {
@@ -181,6 +210,7 @@ export class SoundChip {
         this.volume[2] = volumeTable[v2];
         this.volume[3] = volumeTable[v3];
         this.noisePoked();
+        this._emit({ state: this.snapshotState() });
     }
 
     generate(out, offset, length) {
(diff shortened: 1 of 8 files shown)


mattgodbolt:
On the review's two points: soft resets no longer emit an event, so only a hard reset resyncs the worklet (5c8a48f). The main-thread chip's counters, LFSR and DC state indeed stop advancing in the browser; that is deliberate. Keeping them moving would mean running the generators on the main thread purely for phase, and the only effect of the stale values is that a restore re-syncs the worklet's chip to the snapshot's tone phase and noise position, which is inaudible. Outside the browser (headless harness, tests) the chip has no event sink and renders as before, so its snapshots are unchanged.

src/soundchip.js
```diff
@@ -12,6 +12,10 @@ export const SoundBufferSamples = 512;
 // mean level, which a zero-mean output would silence (see issue #863).
 const DcRestoreCornerHz = 1 / (2 * Math.PI * 10e3 * 4.7e-6);
 
+// A chip with an event sink reports progress this often, so its events can
+// stream through a long execute() rather than all arriving at its end.
+const EventProgressCycles = 4000;
```
mattgodbolt, line 17: Let's think about this, 4000 cycles seems relatively short

src/web/audio-handler.js
```diff
@@ -6,13 +6,13 @@ import { createAudioContext } from "../audio-utils.js";
 import { toggle, fadeIn, fadeOut } from "../dom-utils.js";
 import { toast } from "./toast.js";
 
-// Using this approach means when jsbeeb is embedded in other projects, vite doesn't have a fit.
-// See https://github.com/vitejs/vite/discussions/6459
-const rendererUrl = new URL("./audio-renderer.js", import.meta.url).href;
+// The renderer imports the sound chip, so it must be bundled; a plain URL
+// copies a worklet verbatim (https://github.com/vitejs/vite/discussions/6459).
+import rendererUrl from "./audio-renderer.js?worker&url";
```
mattgodbolt, line 9: if we're sure this is the best way can we fix the music5000 worklet similarly

src/web/audio-renderer.js
```diff
@@ -1,158 +1,189 @@
...
+    // until the producer is a full target ahead again (see _restart).
+    _renderInput(out, length) {
+        if (this.stalled) {
+            this._stall(out, 0, length);
+            return;
+        }
+        let offset = 0;
+        while (length > 0) {
+            const head = this.events[this.eventsHead];
+            if (head !== undefined && isResync(head)) {
```
mattgodbolt: seems maybe this could be simplified by separating the undefined head (presumably this.eventsHead == this.events.size()) case entirely? extract a helper function if there's genuine common code but otherwise it hurts my brain to think about all the head !== undefined stuff

src/soundchip.js
mattgodbolt: The rate buys hiccup tolerance directly: lead oscillates by half the heartbeat spacing, so at 2ms the floor sits about 19ms under a 20ms target, at 5ms about 17.5ms, and on the 10ms tick alone 15ms. The old sample path streamed every 1ms (512 samples). Cost is ~500 small messages a second, most carrying an empty array, against the ~1000 buffers before. I have left it at 2ms; it is one constant if you would rather trade floor for fewer messages.

src/web/audio-handler.js
mattgodbolt: Done: both worklets now import as ?worker&url and the verbatim-copy comment is gone. Build emits both as single IIFEs; checked in vite preview with model=BMusic5000.

src/web/audio-renderer.js
mattgodbolt: Restructured: _samplesUntilNextChange() answers how far the clock can go before the head event or the producer's position (zero for a resync), and _renderInput either renders that far or, at zero, applies the head event or stalls if there is none. The head.cycle > upTo check is gone, since an event can never be past its own batch's position.

PR #908, opened 2026-08-27
fix: drive the PAL subcarrier phase from real scanlines, not texture rows

### LINKED ISSUE #899: PAL sim: texture rows are half-lines, so V-switch and subcarrier phase step at twice the line rate
`src/video.js:905` advances `bitmapY += 2` per HSync (odd interlaced fields start at `bitmapY = -1`), so a 1024x625 texture holds 312.5 scanlines and one Beeb scanline occupies two rows (doubled in the non-interlaced modes that are the default). The shader treats each row as a PAL line: `v_switch = mod(line, 2.0)` and `line * PAL_LINE_PHASE_OFFSET` at `pal-composite.frag.glsl:96` and `:102`.

Consequences:

- Consecutive real scanlines (two rows apart) get a 1.5032-cycle step, i.e. 181 degrees, where real PAL steps 270 degrees (283.7516 cycles per line). The vertical dot and cross-colour pattern has a 2-line period instead of PAL's 4-line one.
- In the doubled modes the two rows of one scanline are encoded with different phase and opposite V-switch, so a single scanline is displayed as two different dot patterns. A shader re-implementation with identical content on rows 10 and 11 (a red/cyan bar edge) differs by 0.67 in RGB at the edge. A real set shows one line, one pattern.
- The 25 Hz offset term (1/625 cycle per line) is applied per row, so doubled.

Suggested fix: derive the line number from the scanline the row belongs to (`floor(row / 2)` or, better, pass the real scanline counter from `video.js` rather than reconstructing it from `gl_FragCoord`), and step phase and V-switch per scanline. Doing this also fixes the per-frame phase step for free once it uses 312 or 312.5 real lines (#901).

The design doc's "outstanding issue 1" identifies a symptom of this but diagnoses it as a 1H versus 2H tap question, which is the wrong diagnosis; the 2H tap already lands on the previous real scanline in every mode.

Confidence: high (source reading plus numeric check).

### LINKED ISSUE #901: PAL sim: per-frame subcarrier phase step is derived from texture rows, not real lines
`PAL_FIELD_PHASE_OFFSET = PAL_LINE_PHASE_OFFSET * PAL_LINES_PER_FIELD` (`pal-composite.frag.glsl:41`, used at `:103`; `pal-composite.js:213`) evaluates to 0.7516 x 312.5 = 0.875 cycles per frame. That is texture-row arithmetic (#899).

Real values: 283.7516 x 312.5 = 88672.375 cycles per interlaced field (fractional part .375), and the Beeb's usual non-interlaced 312-line frame gives 283.7516 x 312 = 88530.4992, fractional part .4992. So on a Beeb the residual dots invert almost every frame (a 25 Hz twinkle) and drift through a full cycle roughly every 50 s. The simulation advances -0.125 cycle per frame instead, a smooth 8-frame (6.25 Hz) crawl.

The doc's "fixed field line count" note (`docs/pal-simulation-design.md:313-317`) covers the interlaced case but not that the constant itself is derived from the wrong quantity. The fix follows from computing phase per real scanline and carrying the true fractional cycle count of the previous frame (312 or 312.5 lines as appropriate) into the next.

Confidence: high on the arithmetic, medium on how noticeable crawl versus twinkle is.

### LINKED ISSUE #903: PAL sim: shader assumes a 1:1 drawing buffer (gl_FragCoord in canvas pixels, TEXTURE_WIDTH hard-coded)
Phase, filter taps and row parity in the shader use `gl_FragCoord` in canvas pixels (`pal-composite.frag.glsl:83`), while the sampling offsets use texel sizes and `TEXTURE_WIDTH = 1024` is hard-coded (`:44`, `:99`). The maths is therefore only right when the drawing buffer is exactly the texture size.

`main.js:2592-2603` scales the backing store by devicePixelRatio only when `maxCanvasScale` is set, which the PAL config does not, so this is latent rather than live today. It will break silently if anyone enables DPR scaling for the PAL filter. Either derive everything from texel coordinates, or pass the real scanline and sample indices in as uniforms/varyings.

Confidence: high.

### LINKED ISSUE #905: PAL sim: add numeric tests for the shader maths
`tests/unit/test-canvas.js` only checks that the PAL program compiles, links and falls back; nothing exercises the maths. Git history (#525, #526, #527) shows precision fixes and FIR reverse-engineering but no numeric validation, which is how the half-line-row phase arithmetic and the mislabelled chroma cutoff went unnoticed.

Suggested tests, runnable in node without WebGL by porting the shader maths (the audit's throwaway node port of the fragment shader took the real `tools/fir-generator.js` coefficients and reproduced the shader well enough to measure everything above):

- Flat colour in, exactly that colour out for all 8 saturated colours (currently passes).
- Identical scanlines decode to identical output regardless of row parity or frame count (currently fails; #899).
- Subcarrier phase advances 283.7516 cycles per real scanline and the V-switch alternates per real scanline.
- The fractional phase carried across a 312-line frame is .4992 and across a 312.5-line field .375.
- The chroma filter's -3 dB point matches whatever the documented figure ends up being.
- The YUV matrix and its inverse round-trip to within float noise.

mattgodbolt:
## What was wrong

The framebuffer texture holds two rows per scanline: `video.js` steps `bitmapY` by 2 per hsync, the usual non-interlaced modes double each scanline onto both rows, and an interlaced field occupies every other row. The PAL composite shader treated every row as a PAL line, so:

- consecutive scanlines were encoded 181 degrees apart rather than PAL's 270 (283.7516 cycles), giving the residual dot pattern a 2-line period instead of 4;
- the two rows of one doubled scanline were encoded with different phase and opposite V-switch, so a single scanline showed two different dot patterns;
- the frame-to-frame phase came from `frameCount % 8` and a 0.7516 x 312.5 constant, a smooth 8-frame crawl that a real Beeb frame (312 lines, 0.4992 cycles) does not produce;
- phase and row parity used `gl_FragCoord` in canvas pixels with a hard-coded 1024-texel line, which only holds while the drawing buffer is exactly the texture size.

## What determines the phase now

`video.js` counts hsyncs in `hsyncCount`, modulo 2500 (0.7516 = 1879/2500, so the phase is exactly periodic and never drifts in float). At each flyback it records the count that texture rows 0 and 1 are about to be drawn under, as `lineBaseEven` and `lineBaseOdd`: the even field's row 0 is the rest of the line the vsync fell in, the odd interlaced field's row 1 is the first hsync after it, and a doubled scanline gives both rows the same line. In interlaced modes the texture holds both fields, so each parity keeps the base from its own field's flyback. The invariant, pinned by tests in every mode: row `r` was drawn under line `lineBase[r & 1] + (r >> 1)`.

The bases travel with each painted frame (`pendingFrame` in `main.js`, `frame` in `canvas.js`) to the filter, which uploads them as `uLineBase` (`vec2`, even then odd). The shader derives texel column and row from `vTexCoord * uResolution`, takes `line` from the base and the row, `v_switch` from the line's parity and `phase = fract(line * 0.7516)`; the previous scanline of the same field is `line - 1`, two rows up. `uFrameCount`, `PAL_FIELD_PHASE_OFFSET` and `TEXTURE_WIDTH` are gone.

Both new fields and the counter are snapshotted; older snapshots restore them as 0.

## What changes on screen

- The dot and cross-colour pattern has PAL's 4-line vertical period.
- A doubled scanline shows one pattern across both of its rows (row pairs are now pixel-identical in MODE 2; before they differed).
- Frame to frame, the residual dots nearly invert each frame (a 25 Hz twinkle drifting through a cycle in about 50 s) in the 312-line modes, and step by 0.375 cycle per 312.5-line interlaced field, rather than the 8-frame crawl.

Verified by eye and numerically with headless Chrome on the MODE 7 boot screen and MODE 2 colour bars.

## Shader tests

`tests/shader/test-pal-composite.js` now drives `uLineBase` through the WebGL harness from #907, and the flat-field cases run at several line numbers rather than frame counts. New cases draw a doubled, cross-coloured MODE 1 picture and check that both rows of a scanline decode identically, that moving it a scanline down decodes differently unless the base is one less (when every row keeps its line number and the picture is identical), that the phase repeats after 2500 lines, and that consecutive lines differ. The whole-picture scale test is unskipped; it compares nearest-sampled renders so that linear sampling's edge softening at two output pixels per texel is not taken for the shader's doing. Against main's shader the doubled-row, line-not-row, consecutive-lines and scale cases fail.

Fixes #899
Fixes #901
Fixes #903
Fixes #905


src/video.js
```diff
@@ -715,9 +743,7 @@ export class Video {
         if (!this.halfClock || !this.oddClock) return;
         if ((this.dispEnabled & EVERYTHINGENABLED) !== EVERYTHINGENABLED) return;
         if (this.bitmapX < 0 || this.bitmapX >= 1024 || this.bitmapY < 0 || this.bitmapY >= 625) return;
-        // The same line doubling decision as the render loop, which inlines it for speed.
-        const doubledLines =
-            (this.doubledScanlines && !this.interlacedSyncAndVideo) || this.isEvenRender === this.lastRenderWasEven;
+        const doubledLines = this.doublesLines();
         const bitmapRow = doubledLines ? this.bitmapY & ~1 : this.bitmapY;
```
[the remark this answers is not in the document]
mattgodbolt: repaintSecondHalfOfCell() runs once per ULA register write (its only caller is Ula.write), not per video tick, so it is not on the hot path; the note it replaced was about the render loop inlining the decision, which it still does. Leaving the method call here.

src/video-filters/shaders/pal-composite.frag.glsl
```diff
@@ -35,13 +34,10 @@ const float PAL_FRAME_RATE = 25.0;           // Frames per second
...
 
 // Derived PAL parameters
-const float PAL_LINES_PER_FIELD = PAL_TOTAL_LINES / 2.0;
 const float PAL_CYCLES_PER_LINE = PAL_SUBCARRIER_MHZ * 1e6 / (PAL_TOTAL_LINES * PAL_FRAME_RATE);
-const float PAL_LINE_PHASE_OFFSET = fract(PAL_CYCLES_PER_LINE);
-const float PAL_FIELD_PHASE_OFFSET = PAL_LINE_PHASE_OFFSET * PAL_LINES_PER_FIELD;
-
-// jsbeeb texture parameters
-const float TEXTURE_WIDTH = 1024.0;          // Framebuffer width (896 visible + 128 blanking)
+// fract(PAL_CYCLES_PER_LINE), as a literal so a float holds it to 1e-8 rather than the 1e-5 it
```
mattgodbolt: what on earth does this comment mean?

src/video-filters/shaders/pal-composite.frag.glsl
```diff
@@ -5,7 +5,8 @@ varying vec2 vTexCoord;
 uniform sampler2D uFramebuffer;
 uniform vec2 uResolution;
 uniform vec2 uTexelSize;
-uniform float uFrameCount;
+// Line number of texture rows 0 (x) and 1 (y); row r is line uLineBase[r & 1] + (r >> 1).
```
mattgodbolt: what on earth does this comment mean

mattgodbolt (commented):
basic lgtm but out of context some of the comments make little sense to me (and/or are overcommenting vs the lines above?)

src/video-filters/shaders/pal-composite.frag.glsl
mattgodbolt: Agreed, it was unreadable. Reworded in e07b8be.

src/video-filters/shaders/pal-composite.frag.glsl
mattgodbolt: Agreed, it was unreadable. Reworded in e07b8be.

PR #924, opened 2026-08-29
feat: put the sound output, speaker amount and display mode on the top bar

mattgodbolt:
Stacked on #923. Puts the settings people will actually want to flip on the top bar, next to play and pause, where they can be changed while the emulator keeps running (the configuration dialog stops it while open):

- **Sound output** (Speaker / Line out / Unfiltered) and the **speaker amount** slider (0 is the board output, 1 the speaker as measured), both live and remembered. The sound output leaves the configuration dialog.
- **Display mode** (RGB Monitor / PAL TV / Smoothed), live, in step with the dialog's own dropdown, and now remembered between visits like the sound settings.

`src/web/quick-settings.js` owns the three controls and does nothing on a page without them (the desktop app strips the bar). Checked in headless Chrome: URL parameters land in the controls, changes apply and persist, the dialog's text follows, no console errors.


mattgodbolt:
Superseded by #928 in the new stack (#926, #927, #928).

PR #946, opened 2026-08-30
ci: run the browser smoke test as test:smoke, in npm test and in CI

mattgodbolt:
Follow-up to #935 now it has been through the Phase 1 PRs.

- `npm run smoke` becomes `npm run test:smoke`, so `npm test`'s `test:*` glob includes it, the same footing as `test:shader`, which already needs Chrome. The pre-commit hook runs only related unit tests, so it does not pick this up, as it does not pick up the integration tests.
- The browser override is `CHROME_PATH`, the name `tests/shader/render.js` already uses.
- The build-and-test job runs it after the shader tests; Chrome is already a requirement of that job.

`CLAUDE.md` updated. Ran locally after the rename: all checks green.

### DIFF
--- a/.github/workflows/test-and-deploy.yml
+++ b/.github/workflows/test-and-deploy.yml
@@ -18,8 +18,9 @@ jobs:
         run: npm install
       - name: checks
         run: npm run ci-checks
-      # Vitest's job summary has a fixed "Vitest Test Report" heading, so each
-      # run gets a heading of its own first to tell the three apart.
+      # Vitest 4's job summary has a fixed "Vitest Test Report" heading, so each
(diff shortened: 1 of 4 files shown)


mattgodbolt:
Added a second commit: each vitest step now writes its own heading (Unit tests, Integration tests, Shader tests) into the job summary before running, since Vitest 4.1's github-actions reporter has a fixed "Vitest Test Report" heading with no option to change it. The smoke test's check lines go into the summary as a list under a heading of their own. Also in this PR: the preview server is bound to 127.0.0.1 and given longer to start, which is what the first CI run needed (localhost resolves to ::1 first on the runner).

Issue #971, opened 2026-08-30
After the split: the next round of design and test improvements


mattgodbolt:
The #638 decomposition was deliberately move-only, so it preserved every wart it met and created a few seams of its own. This is the spec for cleaning those up, from the post-split review (the two outright divergences it found are already fixed by #970, and the pre-existing bugs are #968). Each item names the smallest change that does the job; none is large, and the waves order them so nothing is done twice.

## Wave 1: ownership fixes (each its own small `refactor:` PR)

**1. `MediaLoader` becomes an `EventTarget` and owns `media-changed`.** Done in #986.
Today `setDisc1Image`/`setDisc2Image`/`setTapeImage` dispatch `media-changed` on `config`, purely because `src/app/electron.js` listens there; the settings dialog is acting as a media bus. Move the dispatch onto `MediaLoader` itself, drop its `config` dependency, and have `main.js` forward to electron (or subscribe electron directly). #824's drive readout wants exactly this signal, so do it before #824 starts.

**2. Construct the loop early and pass it, killing the `{isRunning, stop, go}` triple.** Done in #977.
`Modals`, `SnapshotUI` and `RewindUI` each take the same three hand-rolled closures because `EmulationLoop` is built last. Its own dependencies on `frontPanel`/`rewindUI` are already late-bound closures, so it can be built right after machine/display/audio; then the three consumers take `loop` (or, for `Modals`' pause bookkeeping, subscribe to the existing `running` event instead of polling).

**3. Pickers register their fetchers instead of the `sources` bag.** Done in #987.
`main.js` hands `MediaLoader` closures over four objects constructed later. Give the loader `addSource(schema, fetcher)`; each picker registers in its constructor. Removes the forward references and the comment explaining them.

**4. `Layout` becomes a `Settings` wire target; the synthetic `resize` event goes.** Done in #985.
`Display.setMode` currently raises `window` `resize` (immediately and deferred) so that `Layout` relayouts. Wire `layout` into `settings` and call `layout.resize()` directly from `applyDisplayMode`; keep the real window-resize listener, delete both synthetic dispatches (supersedes the immediate one #970 restored).

**5. Small ownership one-liners, one PR:** Done in #978.
- electron's `"save-state"` action calls `snapshots.saveState()` instead of clicking the DOM;
- `hardReset` calls a new `rewindUI.reset()` instead of the close/clear/update triple;
[...]

PR #977, opened 2026-08-31
refactor: build the loop first and hand it over whole

mattgodbolt:
Item 2 of #971.

`Modals`, `SnapshotUI` and `RewindUI` each took the same three hand-rolled `{isRunning, stop, go}` closures because `EmulationLoop` was constructed last. Its own late dependencies (`frontPanel`, `rewindUI`) were already closures, so nothing stopped it being built right after the machine; it moves up with its debug-button listener, and the three consumers now take `loop` itself. Their tests hand in a fake loop instead of three functions; no behaviour changes.

Checked: unit suite, all twelve smoke checks.

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -184,7 +184,7 @@ const quickSettings = new QuickSettings(
 );
 
 // ------------------------------------------------------------------------
-// The machine, and everything that feeds it.
+// The machine, the loop that runs it, and everything that feeds it.
 // ------------------------------------------------------------------------
 
 // Depends on the settings having applied the URL parameters.
@@ -308,7 +308,7 @@ keys.attach({ processor, dbgr, keyLayout });
 const frontPanel = new FrontPanel({ processor, model, printer });
 
 // ------------------------------------------------------------------------
-// Running it: the loop, rewind and the visualiser.
+// Rewind, the visualiser and the layout.
 // ------------------------------------------------------------------------
 
 const rewindUI = new RewindUI({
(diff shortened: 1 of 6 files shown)


src/main.js
```diff
@@ -211,6 +205,34 @@ const machine = new Machine({
...
+    processor,
+    display,
+    audioHandler,
+    dbgr,
+    gamepad,
+    keyboard: keys,
+    syncLights: () => frontPanel.syncLights(),
+    rewindBuffer,
+    onRewindCaptured: () => rewindUI.updateButtonState(),
+    clocksPerSecond,
```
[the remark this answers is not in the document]
mattgodbolt: Deliberate, and unreachable: module evaluation is synchronous, so frontPanel and rewindUI exist before the loop can tick or go() (which first runs in the startup async after machine.start resolves). The loop and rewindUI reference each other, so at least one side must be late-bound; both callbacks staying closures keeps the shape consistent, and a future reorder that broke it would throw loudly on the first tick rather than misbehave.

src/main.js
```diff
@@ -288,40 +311,12 @@ const frontPanel = new FrontPanel({ processor, model, printer });
 // Running it: the loop, rewind and the visualiser.
```
mattgodbolt: "the loop" is not constructed here any more?

mattgodbolt: Right, it rotted the moment the loop moved. Fixed in 0fe27be: the machine section's banner now owns the loop, and the old one names what is actually left (rewind, the visualiser and the layout). A structural banner has to move with the code or it lies; point taken.

Issue #942, opened 2026-08-30
Sound output, speaker amount and display mode picked on the top bar are not written into the URL


mattgodbolt:
## What happens

Picking a sound output, a speaker amount or a display mode with the icon buttons and slider on the top bar (the `#quick-settings` block) takes effect and is remembered in localStorage, but the page URL does not change. Picking the same settings in the Settings dialog does write them into the URL when the dialog closes. Since people share links and reload, a setting chosen on the bar is silently lost from any link copied afterwards.

Observed in headless Chrome against a `vite build` of `main` (5c3c2c4):

1. Fresh load: `http://127.0.0.1:5198/`
2. On the bar: click the board output, drag the speaker amount to 0.5, click PAL. The controls show `board` / `0.5` / `pal` and `localStorage` holds `audioOutput=board`, `speakerAmount=0.5`, `displayMode=pal`, but the URL is still `http://127.0.0.1:5198/`.
3. Open Settings, click the speaker output, close: the URL becomes `http://127.0.0.1:5198/?audioOutput=speaker`.
4. Open Settings, set the speaker amount to 0.25 and click RGB, close: the URL becomes `http://127.0.0.1:5198/?audioOutput=speaker&displayMode=rgb`. Note `speakerAmount` is missing here too (see below).

## Why

In `src/main.js`, `applyAudioOutput`, `applySpeakerAmount` and `applyDisplayMode` are what the top bar's callbacks (wired in the `QuickSettings` constructor in `src/web/quick-settings.js`) call. Each one updates the audio handler or the canvas, mirrors the choice into the dialog and the bar, and writes `window.localStorage`, but none of them touches `parsedQuery` or calls `updateUrl()`.

The dialog path is different. `Config` in `src/config.js` records each click into `this.changed` (`this.changed.audioOutput = ...`, `this.changed.displayMode = ...`) and, when the modal hides, passes it to the `onClose` callback in `src/main.js`, which does `parsedQuery = Object.assign(parsedQuery, changed)` and then `updateUrl()`. So the URL is only ever rewritten by closing the dialog.

The dialog's speaker amount slider has a related gap: its `input` handler in `src/config.js` calls `this.onChange({ speakerAmount })` but never sets `this.changed.speakerAmount`, so even through the dialog the amount never reaches the URL.

## What the fix should be

[...]

PR #990, opened 2026-08-31
refactor: urlState.set applies URL changes in one step (#971 item 7)

mattgodbolt:
Item 7 of #971.

**What changed**

`UrlState` gains `set(changes)`: it applies an object of parameter changes to `params` (a value of `undefined` deletes the key) and pushes the new URL, in one step, so a call site can no longer edit a parameter and forget `updateUrl()` (the #942 class of bug). `params` stays readable everywhere; `updateUrl()` remains public for the debounced case.

Every existing site of the delete/assign/`updateUrl()` ritual migrates:

- `archive-list.js` AutobootTicks (the autoboot tick)
- `analogue-inputs.js` `clearMicrophoneChannel`
- `media-loader.js` local tape load, `setDisc1Image`, `setDisc2Image`, `setTapeImage`, `loadHTMLFile`
- `settings.js` `applyAudioOutput`, `applyDisplayMode`, `onDialogClosed`

**Why behaviour is unchanged**

`set()` performs exactly the edits each call site made by hand, then calls the same `updateUrl()`: one history entry per change, as before. In `onDialogClosed` the merge moves to the top of the method, but nothing between the old and new positions reads the URL, and the code after it reads the same merged `params` (the one visible difference: `Object.assign` left an explicitly undefined `microphoneChannel` as a present-but-undefined key where `set` deletes it; every reader and the URL builder treat those alike). The debounced speaker-amount slider keeps its direct `params` edit because it deliberately defers the URL write so a drag settles into one history entry.

**Tests**

New `UrlState.set` tests, including undefined-deletes-a-key and the pushed URL. The touched modules' tests now exercise a real `UrlState` (with a spied `updateUrl`) instead of a hand-rolled fake, so they cover the new path without weakening any assertion.

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>


src/web/archive-list.js
```diff
@@ -28,12 +28,7 @@ export class AutobootTicks {
...
             check.addEventListener("click", () => {
                 this.show(check.checked);
-                if (check.checked) {
-                    urlState.params.autoboot = "";
-                } else {
-                    delete urlState.params.autoboot;
-                }
-                urlState.updateUrl();
+                urlState.set({ autoboot: check.checked ? "" : undefined });
             });
```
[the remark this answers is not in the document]
mattgodbolt: Confirmed real, but pre-existing: main already sets "" here and the URL builder already dropped it; this PR preserved that behaviour verbatim, as a refactor should. The bug is fixed separately in #993 (with the tests asserting the built URL, which is the gap that let it escape). Once #993 lands this branch will pick up the true value via a merge from main.

src/web/archive-list.js
```diff
@@ -28,12 +28,7 @@ export class AutobootTicks {
         for (const check of this.checks) {
             check.addEventListener("click", () => {
                 this.show(check.checked);
-                if (check.checked) {
-                    urlState.params.autoboot = "";
-                } else {
-                    delete urlState.params.autoboot;
-                }
-                urlState.updateUrl();
+                urlState.set({ autoboot: check.checked ? "" : undefined });
```
mattgodbolt: should this be `true` ? did'nt we fix someting eaerly? either way it seems odd to use empty string for "set" and undefined for "remove"

mattgodbolt: Yes: it is true as of the merge from main (ebac1dd), which picked up #993's fix. The empty string was the pre-existing bug that fix addressed (BOOL params only serialise when exactly true, so the tick never reached the URL); this line now reads urlState.set({ autoboot: check.checked ? true : undefined }). The undefined-removes-the-key convention is set()'s contract, per the PR description.

Issue #1001, opened 2026-08-31
Bugs found by the post-#971 audits


mattgodbolt:
After #971 closed, three read-only audits swept the web layer, the core/web boundary and the test estate. These are the defects they found; each was verified against the source before filing here. The design improvements proposed by the same audits are collected separately in #1002.

All of these except item 1 predate the recent refactoring; the decomposition is what made them findable by reading.

- [x] **1. The CI smoke test step cannot fail.** (#1003) `.github/workflows/test-and-deploy.yml` runs `npm run test:smoke | tee >(...)` with no `shell:` key, so the step runs under plain `bash -e` with no pipefail and the step's exit status is `tee`'s. Every smoke check has been advisory since the suite landed in #995. Fix: `shell: bash`, which makes Actions add `-o pipefail`.

- [x] **2. Save/load state failure resumes the emulator behind the error dialog.** (#1006) `SnapshotUI.saveState` (`src/web/snapshot-ui.js:111`) stops the loop, then on failure opens the error modal; the `Modals` pause latch (`src/web/modals.js:19-26`) samples the already stopped loop, so `savedRunning` latches false, and `saveState`'s own `wasRunning` restarts the loop underneath the dialog. Closing the dialog leaves it running. Two owners hold independent latches for one piece of state; `loadStateFromFile` has the same path, and `RewindUI`'s stop/go pairs should be checked for the same collision.

- [x] **3. A snapshot of the default built-in disc records no way to restore its media, and restoring one can corrupt drive 0.** (#1008) Booting with no `?disc1` picks `BuiltInImages[0]` (`src/main.js:48`) without setting the `disc1` param, so `snapshotMedia` (`src/web/snapshot-ui.js:61-83`) writes `disc1Crc32` and `disc1Layout` but no `disc1` reference. On restore, `reloadSnapshotMedia` (`src/web/snapshot-ui.js:181-195`) finds neither key, loads nothing, and skips the CRC check that exists to catch exactly this; `restoreSnapshot` then lays the snapshot's dirty tracks over whatever is in drive 0. Works by accident when the default disc happens to be loaded again; silently corrupts the drive when the state is loaded into a session started with a different disc.

[...]

PR #1005, opened 2026-08-31
fix: score the debugger's back-disassembly boost on the address, not the text

mattgodbolt:
The debugger's back-disassembly heuristic scores candidate instruction runs leading up to the target address, with a large boost meant for any run that passes through the instruction the CPU actually executed. The boost compared `result[0]` (the disassembled HTML string) against `this.cpu.pc` (a number), so it never fired; back-scrolling relied only on the "common instruction" scoring and could land on a misaligned boundary.

The fix compares the candidate address itself against the program counter. The regression test builds a byte pattern where a misaligned parse (a run of `LDA $41` pairs whose final `LDA $41a9` swallows the real `LDA #$41` at the program counter) outscores the true run unless the boost fires, then checks that scrolling down and back up in the disassembly window returns to the instruction at pc.

Part of #1001 (item 9).

### DIFF
--- a/tests/unit/test-web-debug.js
+++ b/tests/unit/test-web-debug.js
@@ -85,6 +85,21 @@ describe("Debugger", () => {
             expect(currentRow().querySelector(".dis_addr").textContent).toBe("2000");
         });
 
+        it("scrolls back to the instruction at the program counter despite an ambiguous prefix", () => {
+            // Five LDA $41 pairs then $ad give a misaligned parse whose LDA $41a9 swallows
+            // the real LDA #$41 at 2000; only the boost for a run through pc outscores it.
+            for (let i = 0; i < 5; i++) {
+                cpu.writemem(0x1ff5 + i * 2, 0xa5);
+                cpu.writemem(0x1ff6 + i * 2, 0x41);
+            }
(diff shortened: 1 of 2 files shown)


tests/unit/test-web-debug.js
```diff
@@ -85,6 +85,21 @@ describe("Debugger", () => {
             expect(currentRow().querySelector(".dis_addr").textContent).toBe("2000");
         });
 
+        it("scrolls back to the instruction at the program counter despite an ambiguous prefix", () => {
```
mattgodbolt: it's super unfortunate to test the back disassembly we had to fire up the whole UI. Is it feasible to extract the disassembly code in a way that would make this testable, in a relatively small change? If so, do it; else file an issue to track making that change later.

mattgodbolt: Done in 34130e3, the tiny-interface way: prevInstruction is now a free function taking a disassemble callback, the pc and the address; the Debugger method is a one-line delegate. The test builds the real disassembler over a bare 64K array (the opcodes factory only needs peekmem) and never constructs the UI; verified it still fails against the broken comparison. debug.js also turns out to import cleanly in plain node now, so nothing DOM-ish leaks in at module level.

PR #1025, opened 2026-09-01
spike: Playwright for the smoke checks, and what to push down into unit tests

mattgodbolt:
A spike, not a proposal to merge as is. It answers three questions: how much of what `tests/browser/smoke.js` checks belongs in unit tests, what the minimal browser suite is, and whether Playwright should own that suite instead of our own CDP harness. `smoke.js` is untouched and still runs; the Playwright port runs beside it.

## What is on the branch

- `playwright.config.js`, `tests/playwright/fixtures.js`, `tests/playwright/smoke.spec.js`: every `smoke.js` check on `@playwright/test`, plus the top-bar width check from `claude/toolbar-tiers` (marked `fixme` until that lands; main's bar wraps at 1024px). Two projects: `chrome` (the system Chrome, what CI drives today) and `chromium` (the headless shell Playwright pins). `trace: "retain-on-failure"`. The `webServer` is the same `vite preview` of `dist/` on a port of its own. Not part of `npm test`: `npm run e2e`.
- `tests/shader/render.js`: the shader harness hands Chrome to Playwright's library. The `--dump-dom` of the whole document, the regex over it, the title as an error channel and the 256MB stdout buffer go; `page.setContent` and one `page.evaluate` replace them. 32 lines shorter, 6.05s to 4.97s.
- CI: four extra steps (chrome run; cache, shell-only install, chromium run) and a trace upload on failure. Existing steps untouched.

## Part 1: the audit

For each check: (a) unit-testable today with `domFromIndexHtml` and the real modules, (b) unit-testable after a small extraction, (c) needs a real browser. "Already" means the unit test exists on main.

| Check | Class | Seam or extraction | What only the browser catches |
| --- | --- | --- | --- |
| Boots to the BASIC prompt, console surface present | (c), surface (b) | Surface: `exposeConsoleSurface(target, { loop, processor, video, audioHandler })` in `src/web/`, tested against a plain object; the name list then lives at one site. | main.js constructing everything in order against the built bundle, ROMs fetched over HTTP, the machine reaching the prompt. Every other check reuses this. |
| Top bar one row at laptop widths | (c) | none | Layout: Bootstrap CSS, fonts, viewport widths. jsdom does no layout. The CI font fallback (Lato missing) is exactly the kind of thing only this sees. |
| Typing reaches the machine | (a) logic, (c) event shape | Already: `test-keyboard.js`, `test-keyboard-setup.js`. An integration test can dispatch `KeyboardEvent` on `document` with `TestMachine` and read the screen. | The fields Chrome puts on a real `KeyboardEvent` (`keyCode`, `location`) and that main.js wires `KeyboardSetup` at all. |
| Disc named in the URL autoboots | (a) parts, (c) wiring | Already: `test-url-params.js`, `test-web-machine.js`, `test-autoboot.js`. | Only main.js's wiring of those three. Fold into the boot check (boot with `?disc=elite.ssd&autoboot`) rather than keep a page load of its own. |
| A modal pauses and resumes | (a) | Already: `test-modals.js` with synthesised events. Probed here: with `domFromIndexHtml("configuration", "header-bar", ...)`, a click on the nav link opens the real Bootstrap modal through its data API, all four `*.bs.modal` events fire in order, and `Modals` stops and restarts the loop, in jsdom. | The fade transition's timing (`shown.bs.modal` after the CSS transition, hide ignored before it). Keep one modal in the browser for that. |
| Every display mode and sound output picked | (a) bar, (c) effects | Already: `test-quick-settings.js`. | Each mode building a `GlCanvas` and compiling its shader in a real GL context inside the app (the shader tests compile them in a harness, not through `Display`); each output touching WebAudio. Keep. |
| Pause and play buttons | (b) | The `debug-pause`/`debug-play` listeners and the `running` handler in main.js become `RunControls({ loop, dbgr, keys })` in `src/web/`, tested with `domFromIndexHtml("debug-pause", "debug-play")`. | Nothing once extracted. |
| Ctrl-Home into the debugger | (a) | Already: `test-keyboard-setup.js` covers Ctrl-Home to `enterDebugger`. | Nothing. |
| Built-in list into drive 0 | (a) | Already: `test-media-loader.js` lists the images. `utils.loadData` has a Node branch, so an integration test can click the entry and load the real `discs/elite.ssd` into a `TestMachine`'s FDC, then check `urlState`. | Nothing. |
| STH picker opens, lists, closes | (a) | Already: `test-sth-picker.js`. Open and close are the same real-Bootstrap path as the modal probe. The catalogue fetch is external and the check already tolerates its failure, so it proves little. | Nothing the modal check does not. |
| Local disc file | (a) | Already: `test-media-loader.js`, "puts the file in drive 0 and takes the disc out of the URL". | Nothing. |
| Local tape file | (a) | Add the `#tape_load` case to `test-media-loader.js`; the tape parsing is in `test-tapes.js`. | Nothing. |
| Drop on the paste box | (a) | Already: `test-media-loader.js`, "puts a dropped disc in drive 0". The smoke check fakes the `DataTransfer` too, so it is no more real than the unit test. (The Playwright port uses a real `DataTransfer`.) | Nothing. |
| Save state, load it back | (a) | Already: `test-snapshot-ui.js` for the UI; `test-snapshot.js` for the format. Add an integration round trip on `TestMachine`: create, gzip (`CompressionStream` exists in Node), read back, compare screen memory. | `downloadBlob`'s anchor click and the file input, both browser trivia. |
| Shader pixel tests (PAL composite, xBR) | (c) | none | WebGL output. The question is only which harness. |

Minimal browser suite, then: boot with autoboot and the console surface; one real keystroke; one Bootstrap modal; every display mode and sound output; the top bar widths once `toolbar-tiers` lands. Five tests, plus the shaders. Everything else has a headless home.

## Part 2: measurements

Machine: 36 cores, `dist/` already built (a build is 1.7s).

| Harness | Lines | Wall | Notes |
| --- | --- | --- | --- |
| `smoke.js` + `find-chrome.js` (CDP) | 569 + 32 | 17.8s | 13 checks, serial, one Chrome, one page reused. About 190 of those lines are the Chrome launch, the DevTools handshake, the CDP client and the preview server. |
| Playwright, `chrome` project | 39 config + 105 fixtures + 230 spec = 374 | 36.0s / 22.5s / 18.3s at 1 / 2 / 4 workers | 13 checks + 1 fixme. A page and context per test, so each pays a boot. |
| Playwright, `chromium` project | same | 26.5s / 16.2s / 11.8s at 1 / 2 / 4 workers | The headless shell is lighter than Chrome. |
| Playwright, unbounded workers (18) | same | 46s and 3 failures | Real-time emulators starve each other: a held key auto-repeated (`>AAAAB`), cycle counts moved while "paused". The config bounds workers to 4 locally and 2 on CI's 4 vCPUs. |
| Shader suite, `--dump-dom` | render.js 383 | 6.05s | 32 tests, two Chrome invocations |
| Shader suite, Playwright library | render.js 351 | 4.97s | same |

`npx playwright install chromium` took 19.5s here and fetched both the full browser (389MB on disk) and the headless shell (262MB); `--only-shell` fetches the 115MB shell alone, which is all headless runs use. CI step timings are in the section below.

### Diagnosability: the same break on both

I misspelt one selector (`#debug-pause` to `#debug-pausee`) in the pause/play check of each harness.

`smoke.js` (instant):

```
FAIL the pause and play buttons stop and start the emulator
  evaluate failed: Error: No element matches #debug-pausee
```

That is everything. For the CI font failure the whole record was `At 1400px the bar is 152px tall, not 72px`, and finding out why took a font hunt.

Playwright (after the 10s action timeout; before I set `actionTimeout` its auto-waiting burned the whole 60s test timeout, a real cost of the model):

```
TimeoutError: page.click: Timeout 10000ms exceeded.
  - waiting for locator('#debug-pausee')
  > 129 |     await page.click("#debug-pausee");
```

plus `error-context.md` (an accessibility-tree snapshot of the page at the moment of failure, 295 lines, the bar's buttons listed with their labels) and `trace.zip` (626KB): 36 titled steps with parameters and durations (`Navigate to "/"`, `">" on the screen` polls, `Click locator('#debug-pausee')`), 303 screencast frames, 9 DOM snapshots you can hover in the viewer, the 7 console lines, and 26 network exchanges with bodies, the two `woff2` font files among them. Open with `npx playwright show-trace` or trace.playwright.dev; CI uploads `test-results/` on failure. The top-bar failure on main gave the same: the trace shows the wrapped bar at 1024px, which is the picture the font hunt needed.

## Part 3: recommendation

Ranked.

1. **Push down first, whatever happens to the browser layer.** Nine of the fourteen checks are unit-testable today, most already are; the browser copies add only a page load each. Additions: the tape input case; a built-in-list integration test; a save/load round trip on `TestMachine`. Two extractions in the `prevInstruction` spirit: `RunControls` and `exposeConsoleSurface` out of main.js, each a class or function behind a tiny interface, tested against `domFromIndexHtml` or a plain object.
2. **Shrink the browser suite to five tests** (above). This is independent of the harness choice and halves either harness's wall time.
3. **Playwright over the CDP harness for what remains.** Evidence: 374 lines against 601 for the same checks, with no Chrome launch, port handshake, WebSocket client, download interception or file-input plumbing of our own (the day spent on #1022 was on exactly that layer); a trace per failure against one line; real `DataTransfer`, real downloads, real clicks that need the dropdown open (the port opens the Discs and State menus, which `el.click()` via evaluate skipped). Costs, measured: a page per test costs a boot each (18s at 4 workers against 17.8s serial, before the suite shrinks); auto-waiting turns a wrong selector into a timeout, so `actionTimeout` is not optional; a dependency with its own release cadence; the emulator's real-time nature forces a worker bound. Keeping the harness means keeping 200 lines of browser plumbing that have failed twice this month and give nothing when they do.
4. **System Chrome (`channel: "chrome"`) in CI to start.** It needs nothing new: no download, no cache, no `--with-deps`, and it is the browser CI runs now. Switch to the pinned `chromium` project only if a Chrome auto-update on the runner breaks a run; the config already has both, and `--only-shell` keeps the install small. (The font problem is the runner's, not the browser's; either bundle the webfont or keep the check's tolerance for a wider fallback, as `toolbar-tiers` does.)
5. **Cypress: not run here, and not recommended, for reasons I can point at rather than the earlier assessment.** Cypress runs the spec inside the browser beside the app and drives it with synthetic DOM events (its `type()` dispatches `keydown`/`keypress`/`keyup` from script; this is documented, not a guess). For jsbeeb the two checks that justify a browser at all are a real key reaching the matrix and layout at real widths; the first is exactly what synthetic events cannot show, and is what jsdom already gives us. Downloads and `DataTransfer` need plugins or workarounds there; here they are two lines. A one-spec Cypress trial to confirm would need its ~250MB binary and a config; I judged that not worth the spike's budget given the decisive fact is architectural. If you want it run anyway, the keyboard test is the one to run.
6. **vitest browser mode for the shader tests: yes, later, and only that.** The shader harness exists because the tests run in Node: it serialises setup and bind with `Function.prototype.toString` into an HTML string and reads pixels back through base64. In browser mode (`@vitest/browser` with the Playwright provider) the test file itself runs in Chromium, calls WebGL directly, and imports the `.glsl` through vite, so `firShaderPlugin` applies the coefficients the way the app gets them and the manual `applyFirCoefficients` call goes. That is a bigger move than the 32-line change on this branch (a browser project in the vitest config, two more packages, `tools/` stays Node-only), and it is not the right home for the smoke suite: page-level tests want Playwright's own runner (traces, per-test pages, `webServer`). Do it after the smoke suite settles.
7. **CI cost:** see below. The chrome project adds one step and no install; the chromium project adds the install (cached after the first run) and the run.

### Migration, in PRs that each touch one thing

1. `test:` add the tape input unit case, the built-in-list integration test, the save/load round trip. Pure additions.
2. `refactor:` extract `RunControls` and `exposeConsoleSurface` from main.js with unit tests; index.html ids unchanged.
3. `build:` add `@playwright/test`, the config, fixtures and the five-test suite; CI step on `channel: "chrome"`; `smoke.js` stays. (This branch minus the nine redundant tests.)
4. `test:` shader harness onto Playwright's library (second commit here), or straight to vitest browser mode if 6 is wanted now.
5. `test:` remove `smoke.js`, `find-chrome.js`, the smoke CI step and the `test:smoke` script; update CLAUDE.md; add the e2e script to `npm test`.
6. When `toolbar-tiers` lands: un-fixme the top-bar test.

### CI timings

Two runs on ubuntu-24.04 (4 vCPU), step durations from the workflow. The smoke step includes its `npm run build`.

| Step | Run 1 (cold cache) | Run 2 (cache hit) |
| --- | --- | --- |
| Shader tests (`--dump-dom` harness in run 1, Playwright library in both; both commits were in both runs) | 12s | 11s |
| Smoke test (`smoke.js`, build included) | 22s | 24s |
| Playwright smoke, system Chrome (2 workers) | 46s | 44s |
| Cache Playwright's Chromium | miss, 0s | restored, 3s |
| Install Playwright's Chromium (`--with-deps --only-shell`) | 19s | 16s |
| Playwright smoke, bundled Chromium (2 workers) | 28s | 29s |

Two things to read off that. The Playwright port is slower than `smoke.js` on the runner (44s against 24s) because every test boots its own page and 2 workers is all 4 vCPUs will bear; the five-test suite would cut that roughly in proportion, and the bundled headless shell runs the same tests in 29s. And `--with-deps` costs 16s even on a cache hit, since it is apt that takes the time, not the download: on this runner the smoke step already proves Chrome's libraries are present, so the bundled project should install with `--only-shell` alone, or only on a cache miss.

### DIFF
--- a/.github/workflows/test-and-deploy.yml
+++ b/.github/workflows/test-and-deploy.yml
@@ -64,6 +64,29 @@ jobs:
         run: |
           echo "# Smoke test" >> "$GITHUB_STEP_SUMMARY"
           npm run test:smoke | tee >(sed 's/^/- /' >> "$GITHUB_STEP_SUMMARY")
+      # The Playwright spike: the same checks, once in the runner's Chrome and
+      # once in the browser Playwright pins, each timed as its own step.
+      - name: Playwright smoke (system Chrome)
+        run: npx playwright test --project chrome
+      - name: Playwright browser version
+        id: playwright
+        run: echo "version=$(node -p "require('@playwright/test/package.json').version")" >> "$GITHUB_OUTPUT"
+      - name: Cache Playwright's Chromium
+        uses: actions/cache@v4
+        with:
+          path: ~/.cache/ms-playwright
+          key: playwright-chromium-${{ runner.os }}-${{ steps.playwright.outputs.version }}
+      - name: Install Playwright's Chromium
+        run: npx playwright install --with-deps --only-shell chromium
+      - name: Playwright smoke (bundled Chromium)
+        run: npx playwright test --project chromium
+      - name: Keep the Playwright traces of any failure
+        if: failure()
+        uses: actions/upload-artifact@v4
+        with:
+          name: playwright-traces
+          path: test-results/
+          if-no-files-found: ignore
 
   cpu-test:
     runs-on: ubuntu-24.04
--- a/package.json
+++ b/package.json
@@ -36,6 +36,7 @@
   },
   "devDependencies": {
     "@eslint/js": "^10.0.1",
+    "@playwright/test": "^1.62.1",
     "@vitest/coverage-v8": "^4.0.18",
     "eslint": "^10.7.0",
     "eslint-config-prettier": "^10.1.8",
@@ -115,7 +116,8 @@
     "mirror-bbcdiscs:upload:index": "aws s3 cp .bbcdiscs-mirror/hfe/manifest.json s3://bbc.xania.org/archive/bbcdiscs/hfe/manifest.json --cache-control 'public, max-age=300' && aws s3 cp .bbcdiscs-mirror/manifest.json s3://bbc.xania.org/archive/bbcdiscs/manifest.json --cache-control 'public, max-age=300'",
     "electron": "npm run build && ELECTRON_DISABLE_SANDBOX=1 electron .",
     "electron:build": "electron-builder",
-    "test:smoke": "npm run build && node tests/browser/smoke.js"
+    "test:smoke": "npm run build && node tests/browser/smoke.js",
+    "e2e": "npm run build && playwright test"
   },
   "lint-staged": {
     "*.js": [
--- a/playwright.config.js
+++ b/playwright.config.js
@@ -1,0 +1,42 @@
+import { defineConfig } from "@playwright/test";
+
+// The preview of dist/ the tests drive. A port of its own, so this and
+// tests/browser/smoke.js can run side by side.
+const PreviewHost = "127.0.0.1";
+const PreviewPort = 5198;
+const PreviewUrl = `http://${PreviewHost}:${PreviewPort}/`;
+
+export default defineConfig({
+    testDir: "tests/playwright",
+    fullyParallel: true,
+    forbidOnly: !!process.env.CI,
+    workers: process.env.CI ? 2 : undefined,
+    reporter: process.env.CI ? [["list"], ["github"]] : "list",
+    timeout: 60000,
+    expect: { timeout: 10000 },
+    use: {
+        // Without this a click on a selector that matches nothing waits out
+        // the whole test timeout.
+        actionTimeout: 10000,
+        baseURL: PreviewUrl,
+        viewport: { width: 1400, height: 900 },
+        trace: "retain-on-failure",
+        launchOptions: {
+            // Playwright's own headless switches already mute audio and keep
+            // shared memory out of /dev/shm; these are what the app needs on top.
+            args: ["--use-gl=angle", "--use-angle=swiftshader", "--autoplay-policy=no-user-gesture-required"],
+        },
+    },
+    projects: [
+        // The system Chrome, as the CDP harness and CI use today.
+        { name: "chrome", use: { channel: "chrome" } },
+        // The build Playwright ships and pins to its own version.
+        { name: "chromium" },
+    ],
+    webServer: {
+        // vite's own entry point, not npx, so the kill reaches the server itself.
+        command: `node node_modules/vite/bin/vite.js preview --host ${PreviewHost} --port ${PreviewPort} --strictPort`,
+        url: PreviewUrl,
+        reuseExistingServer: !process.env.CI,
+    },
+});
--- a/playwright.config.js
+++ b/playwright.config.js
@@ -10,7 +10,10 @@ export default defineConfig({
     testDir: "tests/playwright",
     fullyParallel: true,
     forbidOnly: !!process.env.CI,
-    workers: process.env.CI ? 2 : undefined,
+    // Each page runs the machine in real time, so the checks that watch it
+    // stop and start need CPU of their own: with one worker per core they
+    // starve each other and keys auto-repeat.
+    workers: process.env.CI ? 2 : 4,
     reporter: process.env.CI ? [["list"], ["github"]] : "list",
     timeout: 60000,
     expect: { timeout: 10000 },
--- a/tests/playwright/fixtures.js
+++ b/tests/playwright/fixtures.js
@@ -44,8 +44,12 @@ class Beeb {
             .not.toContain(text);
     }
 
+    // currentCycles wraps every emulated second; the seconds term keeps this monotonic.
     cycles() {
-        return this.page.evaluate(() => window.processor.currentCycles);
+        return this.page.evaluate(
+            () =>
+                window.processor.cycleSeconds * window.processor.model.cyclesPerSecond + window.processor.currentCycles,
+        );
     }
 
     /** Checks the emulator stays where it is for a while, and returns where that is. */
--- a/tests/playwright/fixtures.js
+++ b/tests/playwright/fixtures.js
@@ -1,0 +1,105 @@
+import { expect, test as base } from "@playwright/test";
+
+const ScreenBase = 0x7c00;
+const ScreenBytes = 1000;
+const BootTimeoutMs = 30000;
+const KeyHoldMs = 120;
+
+/** The page with jsbeeb on it, reached through the console surface main.js exposes. */
+class Beeb {
+    constructor(page) {
+        this.page = page;
+        this.problems = [];
+        page.on("pageerror", (error) => this.problems.push(`exception: ${error.message}`));
+        page.on("console", (message) => {
+            if (message.type() === "error") this.problems.push(`console error: ${message.text()}`);
+        });
+    }
+
+    open(query = "") {
+        return this.page.goto(`/${query}`);
+    }
+
+    /** Mode 7 screen memory as text, which is what the machine shows at the prompt. */
+    screenText() {
+        return this.page.evaluate(
+            ([base, bytes]) => {
+                const cpu = window.processor;
+                if (!cpu) return "";
+                return Array.from({ length: bytes }, (_, i) => String.fromCharCode(cpu.readmem(base + i) & 0x7f)).join(
+                    "",
+                );
+            },
+            [ScreenBase, ScreenBytes],
+        );
+    }
+
+    async expectScreenText(text, timeout = BootTimeoutMs) {
+        await expect.poll(() => this.screenText(), { message: `"${text}" on the screen`, timeout }).toContain(text);
+    }
+
+    async expectNotOnScreen(text, timeout = BootTimeoutMs) {
+        await expect
+            .poll(() => this.screenText(), { message: `"${text}" to leave the screen`, timeout })
+            .not.toContain(text);
+    }
+
+    cycles() {
+        return this.page.evaluate(() => window.processor.currentCycles);
+    }
+
+    /** Checks the emulator stays where it is for a while, and returns where that is. */
+    async expectHeld(ms = 300) {
+        const before = await this.cycles();
+        await this.page.waitForTimeout(ms);
+        expect(await this.cycles(), "cycles run while the emulator should be stopped").toBe(before);
+        return before;
+    }
+
+    async expectRunningPast(cycles) {
+        await expect.poll(() => this.cycles(), { message: "the emulator to run again" }).toBeGreaterThan(cycles);
+    }
+
+    /** Held long enough for the OS to scan the matrix and see it down. */
+    async pressKey(key) {
+        await this.page.keyboard.down(key);
+        await this.page.waitForTimeout(KeyHoldMs);
+        await this.page.keyboard.up(key);
+    }
+
+    drive0() {
+        return this.page.evaluate(() => window.processor.fdc.drives[0].disc?.name ?? null);
+    }
+
+    async expectDrive0(name, timeout = BootTimeoutMs) {
+        await expect.poll(() => this.drive0(), { message: `${name} to be in drive 0`, timeout }).toBe(name);
+    }
+
+    /**
+     * Bootstrap ignores a hide while its show transition is still running, and
+     * says when that is over. Call before the click that opens the modal.
+     */
+    modalShown(id) {
+        return this.page.evaluate(
+            (modalId) =>
+                new Promise((resolve) =>
+                    document.getElementById(modalId).addEventListener("shown.bs.modal", resolve, { once: true }),
+                ),
+            id,
+        );
+    }
+
+    forgiveProblems(pattern) {
+        this.problems = this.problems.filter((problem) => !pattern.test(problem));
+    }
+}
+
+export const test = base.extend({
+    beeb: async ({ page }, use) => {
+        const beeb = new Beeb(page);
+        await use(beeb);
+        expect(beeb.problems, "errors the page logged").toEqual([]);
+    },
+});
+
+export { expect };
--- a/tests/playwright/smoke.spec.js
+++ b/tests/playwright/smoke.spec.js
@@ -1,0 +1,230 @@
+// The checks of tests/browser/smoke.js, on Playwright: what only the built
+// page in a real browser can show. Each test gets a page of its own.
+//
+//   npm run e2e                       build, then run every check in the system Chrome
+//   npx playwright test --project chromium   in the browser Playwright ships
+//   npx playwright test -g modal             checks whose name matches
+//   npx playwright show-trace test-results/**/trace.zip
+
+import path from "node:path";
+
+import { expect, test } from "./fixtures.js";
+
+const ConsoleSurface = [
+    "processor",
+    "video",
+    "soundChip",
+    "go",
+    "stop",
+    "hd",
+    "m7dump",
+    "benchmarkCpu",
+    "profileCpu",
+    "benchmarkVideo",
+    "profileVideo",
+];
+
+const OneRowWidth = 4000;
+const LaptopWidths = [1024, 1280, 1400, 1440, 1512, 1536, 1600];
+
+test("boots to the BASIC prompt with a working console surface", async ({ beeb, page }) => {
+    await beeb.open();
+    await beeb.expectScreenText("BASIC");
+    await beeb.expectScreenText(">");
+    const missing = await page.evaluate(
+        (names) => names.filter((name) => typeof window[name] === "undefined"),
+        ConsoleSurface,
+    );
+    expect(missing, "names missing from window").toEqual([]);
+    expect(typeof (await page.evaluate(() => window.processor.readmem(0)))).toBe("number");
+});
+
+test("the top bar is one row at laptop widths", async ({ beeb, page }) => {
+    test.fixme(true, "main's bar wraps below 1400px; the fix is on claude/toolbar-tiers, whose smoke check this ports");
+    const headerBar = async (width) => {
+        await page.setViewportSize({ width, height: 900 });
+        return page.evaluate(() => {
+            const bar = document.getElementById("header-bar");
+            const paste = document.getElementById("paste-text").getBoundingClientRect();
+            const promo = bar.querySelector(".navbar-text");
+            const promoState = () => {
+                if (getComputedStyle(promo).display === "none") return "hidden";
+                if (promo.getBoundingClientRect().right > promo.nextElementSibling.getBoundingClientRect().left)
+                    return "overflowing";
+                return promo.scrollWidth > promo.clientWidth ? "truncated" : "fits";
+            };
+            return {
+                height: bar.offsetHeight,
+                pasteFits: paste.width > 0 && paste.right <= innerWidth,
+                promo: promoState(),
+            };
+        });
+    };
+    await beeb.open();
+    await beeb.expectScreenText(">");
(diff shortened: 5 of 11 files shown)


tests/playwright/fixtures.js
```diff
@@ -0,0 +1,105 @@
...
+    /** Checks the emulator stays where it is for a while, and returns where that is. */
+    async expectHeld(ms = 300) {
+        const before = await this.cycles();
+        await this.page.waitForTimeout(ms);
+        expect(await this.cycles(), "cycles run while the emulator should be stopped").toBe(before);
+        return before;
+    }
+
+    async expectRunningPast(cycles) {
+        await expect.poll(() => this.cycles(), { message: "the emulator to run again" }).toBeGreaterThan(cycles);
```
mattgodbolt, line 64: **This is the port's one real bug, and it fails runs frequently.** `processor.currentCycles` is not monotonic: src/6502.js:1499-1504 subtracts `cyclesPerSecond` from it every emulated second and increments `cycleSeconds`. So after resume the counter can sit below `cycles` for most of each emulated second, and `toBeGreaterThan` only succeeds if a poll sample happens to land in the window above the saved value. expect.poll samples about 13 times in its 10s budget (100/250/500/1000ms intervals), so when the pause landed near the top of the 2,000,000 range the test usually fails. Measured: 3 of 7 full chrome-project runs at this config's default 4 workers on an idle 36-core machine failed here (modal once, pause/play twice, both in one run). A captured failure shows the shape exactly: `Expected: > 1934174, Received: 1635398` after 10 seconds of a machine that was demonstrably running.

smoke.js has the same latent hazard (its waitUntil polls the same expression) but at 10Hz it gets about 100 samples and looks stable; roughly 1 in 100 runs per affected check would still miss.

Fix in whichever harness survives, before any suite shrink (the minimal five-test suite keeps a modal test on this helper): poll a monotonic value, e.g. `cycleSeconds * cyclesPerSecond + currentCycles`, or simply assert that the value changes rather than grows.

playwright.config.js
```diff
@@ -0,0 +1,45 @@
...
+const PreviewUrl = `http://${PreviewHost}:${PreviewPort}/`;
+
+export default defineConfig({
+    testDir: "tests/playwright",
+    fullyParallel: true,
+    forbidOnly: !!process.env.CI,
+    // Each page runs the machine in real time, so the checks that watch it
+    // stop and start need CPU of their own: with one worker per core they
+    // starve each other and keys auto-repeat.
+    workers: process.env.CI ? 2 : 4,
```
mattgodbolt, line 16: The bound is justified (the 18-worker starvation in the PR body is real), but note the failures I saw at this default of 4 on a 36-core machine were not starvation; they were the cycle-counter wrap (see fixtures.js:60). Once that is fixed, decide `retries` deliberately: there is none configured, and a real-time emulator on a shared 4 vCPU runner will still have a tail. `retries: process.env.CI ? 1 : 0` with the existing retain-on-failure trace would be the conventional shape if any is wanted.

package.json
```diff
@@ -115,7 +116,8 @@
     "mirror-bbcdiscs:upload:index": "aws s3 cp .bbcdiscs-mirror/hfe/manifest.json s3://bbc.xania.org/archive/bbcdiscs/hfe/manifest.json --cache-control 'public, max-age=300' && aws s3 cp .bbcdiscs-mirror/manifest.json s3://bbc.xania.org/archive/bbcdiscs/manifest.json --cache-control 'public, max-age=300'",
     "electron": "npm run build && ELECTRON_DISABLE_SANDBOX=1 electron .",
     "electron:build": "electron-builder",
-    "test:smoke": "npm run build && node tests/browser/smoke.js"
+    "test:smoke": "npm run build && node tests/browser/smoke.js",
+    "e2e": "npm run build && playwright test"
```
mattgodbolt, line 120: `playwright test` with no `--project` runs both projects (28 tests, verified with --list), so `npm run e2e` contradicts the spec header's own description of it (build, then run every check in the system Chrome) and half-fails on any machine without the pinned Chromium: verified with PLAYWRIGHT_BROWSERS_PATH pointed at an empty directory, the chrome half passes and the chromium half errors. A fresh contributor after npm ci gets 13 passes and 13 browser-launch failures. Either `playwright test --project chrome` here, or document the install step; the chromium project stays reachable by hand either way.

.github/workflows/test-and-deploy.yml
```diff
@@ -64,6 +64,29 @@ jobs:
         run: |
           echo "# Smoke test" >> "$GITHUB_STEP_SUMMARY"
           npm run test:smoke | tee >(sed 's/^/- /' >> "$GITHUB_STEP_SUMMARY")
+      # The Playwright spike: the same checks, once in the runner's Chrome and
+      # once in the browser Playwright pins, each timed as its own step.
+      - name: Playwright smoke (system Chrome)
+        run: npx playwright test --project chrome
```
mattgodbolt, line 70: This step only works because the Smoke test step above it ran `npm run build` a moment earlier; `npx playwright test` does not build and the webServer only serves dist. Fine while smoke.js stays, but migration step 5 deletes that step, which orphans these: nothing in the job would build dist any more (the unit/integration/shader steps do not). When smoke.js goes, either switch this to `npm run e2e` or add an explicit build step.

.github/workflows/test-and-deploy.yml
```diff
@@ -64,6 +64,29 @@ jobs:
...
+      - name: Playwright browser version
+        id: playwright
+        run: echo "version=$(node -p "require('@playwright/test/package.json').version")" >> "$GITHUB_OUTPUT"
+      - name: Cache Playwright's Chromium
+        uses: actions/cache@v4
+        with:
+          path: ~/.cache/ms-playwright
+          key: playwright-chromium-${{ runner.os }}-${{ steps.playwright.outputs.version }}
+      - name: Install Playwright's Chromium
+        run: npx playwright install --with-deps --only-shell chromium
```
mattgodbolt, line 80: The PR body's own CI analysis concludes `--with-deps` costs 16s even on a cache hit and that the smoke step already proves the runner has Chrome's libraries; the branch should match its own recommendation (`--only-shell` alone, or deps only on a cache miss). As committed, every push pays the apt run.

tests/playwright/smoke.spec.js
```diff
@@ -0,0 +1,230 @@
...
+    const missing = await page.evaluate(
+        (names) => names.filter((name) => typeof window[name] === "undefined"),
+        ConsoleSurface,
+    );
+    expect(missing, "names missing from window").toEqual([]);
+    expect(typeof (await page.evaluate(() => window.processor.readmem(0)))).toBe("number");
+});
+
+test("the top bar is one row at laptop widths", async ({ beeb, page }) => {
+    test.fixme(true, "main's bar wraps below 1400px; the fix is on claude/toolbar-tiers, whose smoke check this ports");
```
mattgodbolt, line 43: `test.fixme(true, ...)` aborts at runtime, so everything below is dead code until claude/toolbar-tiers lands and someone remembers to delete this line; meanwhile the selectors it reaches (`#paste-text`, `.navbar-text`, the promo's sibling) can drift with that branch and nothing will notice. A skipped test that can rot is spike-acceptable, but the un-fixme step in the migration plan should include re-validating the body against what toolbar-tiers actually merged, not just removing the marker.

tests/shader/render.js
```diff
@@ -297,47 +289,23 @@ document.title = "OK";
...
  */
-export function renderJobs(filter, jobs) {
-    const chrome = findChrome();
-    const dir = mkdtempSync(path.join(tmpdir(), "jsbeeb-shader-"));
-    const pageFile = path.join(dir, "harness.html");
-    writeFileSync(pageFile, buildHarness(filter, jobs));
+export async function renderJobs(filter, jobs) {
+    // The system Chrome, as the smoke test's "chrome" project drives.
+    const browser = await chromium.launch({
+        channel: "chrome",
```
mattgodbolt, line 298: This narrows browser discovery. find-chrome.js honoured CHROME_PATH and fell back through google-chrome, google-chrome-stable, chrome, chromium and chromium-browser; `channel: "chrome"` finds only branded Google Chrome. A Linux box with only Chromium installed could run the shader tests before this change and cannot after, and CHROME_PATH silently stops working. Suggest `executablePath: process.env.CHROME_PATH` when set, else the channel, else the pinned browser as the smoke suite's chromium project already does.

mattgodbolt (commented):
Independent review of the spike, with everything below measured on this branch (36 cores, same class of machine as the PR body's measurements). Summary of what I verified and what I found; the file-anchored findings are inline.

## What checks out

- **Fidelity**: all 13 smoke.js checks have a Playwright counterpart asserting at least as much. Several are stronger: the port opens the real Discs and State menus where smoke.js clicks hidden elements from script, uses a real DataTransfer for the drop test, and a real download event for save state. I proved sensitivity with three deliberate breaks in dist/index.html (debug-pause id renamed, the configuration nav link retargeted, tape_load id renamed): both harnesses catch all three (two of them crash main.js at startup, so every check fails in both).
- **The jsdom modal claim**: I wrote an independent probe (jsdom, domFromIndexHtml with header-bar and configuration, real Modals, real Bootstrap data API): the nav link click opens the modal, all four *.bs.modal events fire in order, and Modals stops and restarts the loop. The audit's claim is true.
- **Timings**: smoke.js 17.6s (claimed 17.8s); chrome project 19.6s at 4 workers, 21.9s and 22.1s at 2 (claimed 18.3s / 22.5s, low variance); chromium project 10.7s at 4 workers (claimed 11.8s); shader harness 5.94s on main's --dump-dom version against 4.93s on the Playwright version (claimed 6.05s / 4.97s), 32 tests passing both ways with identical assertions (the test files change only by an await). The CI timing table matches the actual run 33537006183 step for step (44s / 3s / 16s / 29s).
- **System Chrome needs nothing installed**: verified; the chrome project passes with PLAYWRIGHT_BROWSERS_PATH pointed at an empty directory. The chromium project fails without an install, which feeds the e2e-script finding inline.

## The one big problem

The port is much flakier than the harness it replaces, and not for the starvation reason the config guards against. In 7 full chrome-project runs at the config's own local default of 4 workers on an otherwise idle 36-core machine, 3 runs had failures (modal once, pause/play twice, both once). Root cause is inline on fixtures.js:60: `processor.currentCycles` wraps every emulated second (src/6502.js:1499-1504 subtracts cyclesPerSecond and increments cycleSeconds), so `toBeGreaterThan(stopped)` can only succeed while the counter is above the wrapped `stopped` value, and expect.poll's sparse sampling (about 13 samples in 10s) misses that window when `stopped` lands near the top of the 2,000,000 range. My captured failure shows exactly this: Expected > 1934174, Received 1635398 after 10s of a running machine. smoke.js has the same latent hazard but polls at 10Hz (about 100 samples), which is why it looks stable; back of the envelope it still fails about 1 in 100 runs per affected check. The fix (compare cycleSeconds * cyclesPerSecond + currentCycles, or the tuple) belongs in whichever harness survives, and should land before any suite shrink, since the minimal five-test suite keeps a modal test built on the same helper.

## Migration plan gaps

- Step 5 removes smoke.js and its CI step, but the CI Playwright steps run `npx playwright test` bare and the webServer only serves dist; today the build they depend on comes from the smoke step they would be deleting (inline on the workflow).
- Folding autoboot into the boot check (audit table, first two rows) drops the plain no-parameter boot path from the browser suite; the default-config machine reaching the BASIC prompt would no longer be covered in a browser.
- smoke.js's `--url` mode goes with it, and CLAUDE.md documents running against a dev server on 5173; the port pins baseURL to 5198. Worth an env override for baseURL before step 5.
- No `retries` are configured; that may be the right call once the wrap fix lands, but it should be a decision, not an omission.

## Small claim rot

The body says 39 config lines and 374 total; the last commit grew the config to 45 (380 total). 601 for smoke.js + find-chrome.js, 105 fixtures, 230 spec, and 383 to 351 for render.js all check out.

Everything else I looked for and did not find: no port collisions (5198 preview, 5199 smoke, 4173/5173 vite defaults all distinct), vitest exclusion of tests/playwright is correct and npm test picks up no spec files, lint passes, trace upload fires on step failure and if-no-files-found keeps it quiet otherwise, and the chromium runs in CI do not wipe a failed chrome run's traces because a chrome failure skips them.

tests/playwright/fixtures.js
```diff
@@ -0,0 +1,105 @@
...
+    /**
+     * Bootstrap ignores a hide while its show transition is still running, and
+     * says when that is over. Call before the click that opens the modal.
+     */
+    modalShown(id) {
+        return this.page.evaluate(
+            (modalId) =>
+                new Promise((resolve) =>
+                    document.getElementById(modalId).addEventListener("shown.bs.modal", resolve, { once: true }),
+                ),
```
mattgodbolt, line 91: Minor: when a modal test fails before the event fires, this evaluate's promise is left dangling and Playwright reports an extra `Error: page.evaluate: Test ended.` alongside the real failure (observed when I broke the configuration nav link). smoke.js's flag-plus-poll shape avoids the noise; installing the listener in one short evaluate and polling a window flag would keep the report clean here too.

mattgodbolt:
The review's finding 1 (expectRunningPast polling the wrapping currentCycles) is fixed on this branch in 00cad83: the fixtures now poll cycleSeconds * model.cyclesPerSecond + currentCycles. Verified with the three affected tests repeated six times each at 4 workers, all passing; the same fix for smoke.js on main is #1026. One wrinkle worth recording: cyclesPerSecond lives on the model, not the processor, and the wrong form evaluates to NaN in the page, which turns the flake into a deterministic timeout.

mattgodbolt:
Superseded by #1030, which landed the production version of this spike (the minimal suite, smoke.js deleted, every check homed). Kept for its audit and review record.

PR #1033, opened 2026-09-02
feat: track executed instructions in the debugger disassembly

### LINKED ISSUE #1016: Back-disassembly heuristic: measured misses and what to do about them
Research into how debuggers find the previous instruction when scrolling backwards through a variable-length instruction stream, and how jsbeeb's `Disassemble6502.prevInstruction` (src/6502.opcodes.js) measures up. Follow-on from #1005 and #1013.

## What everyone else does

Nobody has a clever answer. Every mature tool starts some distance back, disassembles forward and keeps the parse that lands on the target, leaning on side information when it has any (symbols, analysis databases, execution logs):

- GDB's TUI starts at the nearest symbol before the address and iterates; without symbols it gives up ("There is no safe way how to 'disassemble backwards'"). [tui-disasm.c](https://course.khoury.northeastern.edu/cs3650/parent/gdb-12-src/gdb-12.1/gdb/tui/tui-disasm.c)
- x64dbg starts 16 * (n + 3) bytes back, records instruction starts in a ring and takes the n-th before ip, without checking it lands exactly; the maintainer calls it "slightly tweaked from OllyDbg". [disasm_helper.cpp](https://raw.githubusercontent.com/x64dbg/x64dbg/development/src/dbg/disasm_helper.cpp), [thread](https://groups.google.com/g/x64dbg/c/R2AFihikB9U)
- MAME starts 3 * 64 bytes back and, if the forward pass misses pc, shows nothing above it. [dvdisasm.cpp](https://raw.githubusercontent.com/mamedev/mame/master/src/emu/debug/dvdisasm.cpp)
- radare2 asks analysis for the containing basic block first, then "fallback to the dumb solution": 32 bytes back, decode forward. [visual.c](https://raw.githubusercontent.com/radareorg/radare2/master/libr/core/visual.c)
- Ghidra keeps code units in a database, so previous is a lookup (`Listing.getInstructionBefore`). [API](https://ghidra.re/ghidra_docs/api/ghidra/program/model/listing/Listing.html)
- Mesen (the 6502-family emulator precedent) logs which bytes were fetched as opcodes, marks lines verified vs unexecuted, and realigns when a verified start appears inside a guessed instruction. [Disassembler.cpp](https://raw.githubusercontent.com/SourMesen/Mesen2/master/Core/Debugger/Disassembler.cpp), [docs](https://www.mesen.ca/docs/debugging/debugger.html)
- b2 steps back 1, 2 or 3 bytes based on the opcode found there, with no resynchronisation check at all. [debugger.cpp](https://raw.githubusercontent.com/tom-seddon/b2/master/src/b2/debugger.cpp)

The theory (Linn and Debray's "self-repairing" disassembly, [CCS 2003](https://www2.cs.arizona.edu/~debray/Publications/disasm-resist.pdf); Miller et al., [ICSE 2019](https://yonghwi-kwon.github.io/data/probdisasm_icse19.pdf)): a wrong start emits garbage whose lengths are effectively random relative to the true boundaries, so it rejoins the real stream quickly and stays there. Measured on the 6502 with jsbeeb's disassembler: a parse started one or two bytes inside a real instruction resynchronises within 1 instruction 48% of the time, within 2 71%, within 4 88%, within 9 96%. So a window of 20 bytes is marginal and 64 is comfortable, and the choice among candidate runs rarely matters; the only real ambiguity is where the bytes before the target are data or unexecuted code, which is about 1 scroll step in 16 on the MOS.

## How jsbeeb's heuristic measures

Ground truth: 6074 distinct executed instruction starts in MOS, BASIC and DFS from a Model B boot plus a short BASIC session; pc placed outside the window so the +10 boost is inert.

| Variant | Accuracy |
|---|---|
| current, window 20 | 99.6% (22 misses of 5689) |
| current, window 64 | 99.8% |
| `bestScore` initialised to -1 and non-legal opcodes penalised, window 20 | 99.8% |
| same, window 64 | 99.9% |
| earliest resynchronising run, no scoring at all, window 64 | 99.8% |
| log-probability scoring from a frequency table, window 64 | 99.8% |

16 of the 22 misses are one bug: the unrolled `STA $xx00,X` screen clear at &CC63 to &CCDD. Every instruction there is excluded by the "uncommon indexed mode" regex, so the true run scores 0, `score > bestScore` never fires against the initial 0, and the default `address - 1` (a bogus instruction made from the address high byte) wins. Initialising `bestScore` to -1 (any run that lands on the target beats none) fixes all 16. Most of the rest are junk parses through undocumented opcodes (ISB, SRE, SAX, `???`) next to padding; penalising the 105 non-legal opcodes fixes those. Static opcode frequency on the same corpus also shows `STA abs,X` in the top ten, so the indexed-mode exclusion is wrong for 6502 code in general and should at most be a tie-breaker.

## What the emulator can do that a static tool cannot

`Cpu6502.execute` already switches from `executeInternalFast` to the hooked `executeInternal` whenever a debug hook is installed, so a `debugInstruction` hook writing `executed[pc] = opcode + 1` costs nothing with the debugger closed and was within run-to-run noise with it open (40M cycles headless: 1.63 to 1.75 s unhooked vs 1.56 to 2.00 s hooked). Storing the opcode byte rather than a flag makes the map self-validating (an entry is trustworthy exactly when `executed[a] === peekmem(a) + 1`, since a 6502 instruction's length depends only on its opcode), which covers self-modifying code and overwritten RAM with no write hook; sideways ROM needs the 0x8000 to 0xBFFF range keyed by bank as well. With the map, scrolling back is exact through executed code, and it enables the Mesen presentation: dim never-executed lines and show byte runs between executed instructions as data rather than fake instructions.

## Recommendations, ranked

1. **Initialise `bestScore` to -1** in `prevInstruction`. One line, no hot-path cost, removes 16 of 22 measured misses. Unit test on a run of `9D 00 xx`.
2. **Penalise non-legal opcodes and widen the window to 64.** Same function, tens of lines; the disassembler table already names the undocumented ops. Drop the indexed-mode exclusion or demote it to a tie-breaker. With (1): 99.9%. Cost is O(window^2) `disassemble` calls per scroll step (about 1000 at 64 bytes), fine per wheel notch; a length-only decode would make it negligible.
3. **Executed-opcode map while the debugger is enabled**, plus an "exact if known" first step in `prevInstruction` and per-bank storage for sideways ROM. Touches src/6502.js or src/web/debug.js (install and remove the hook in `Debugger.enable`). Its value is mostly the presentation win, since (1)+(2) already reach 99.9% where an answer exists.
4. **Cheap anchors** (the existing `oldPcArray`, the MOS vectors at 0x200 to 0x235) treated like the pc boost. Small, and mostly redundant once (3) exists.
5. **Not worth it**: frequency-weighted scoring, branch-target consistency bonuses, recursive descent from entry points. None measured better than (1)+(2) on this corpus.

Caveats: one boot session of MOS 1.20, BASIC 2 and DFS 1.2 with a short BASIC program; game code was not measured (Repton 2 at &2CBB, the comment's own test case, was not available in the run). MOS &CFC8 already resolves correctly.

mattgodbolt:
Item 3 of #1016: the executed-opcode map, exact back-disassembly through executed code, and the Mesen-style presentation. Closes #1016.

## What it does

`ExecutionMap` (src/execution-map.js) records `opcode + 1` for every fetched instruction start, zero meaning never seen. Entries vouch for themselves: one is trusted exactly when `map[addr] === peekmem(addr) + 1`, so self-modifying code and overwritten RAM invalidate themselves with no write hook. The sideways region is keyed by `romsel & 15` as well, since each bank has its own code at the same addresses; Master shadow and private RAM alias that keying, and the opcode compare catches nearly all of it, as the issue sketched.

The recording hook is added in `Debugger.enable(true)` and removed in `enable(false)`. `Cpu6502.execute` picks the hooked loop only while a debug hook exists, so with the debugger closed the cost is exactly zero; a unit test asserts the CPU is back on `executeInternalFast` after the debugger is hidden.

`prevInstruction` gains an optional third argument: with a map, a verified executed instruction ending exactly at the target is returned before any guessing, and with no map nothing changes. The debugger passes its map through; that one optional parameter is the whole seam.

In the disassembly window, rows whose address is not a verified executed instruction start get an `unexecuted` class and render dimmed, so real code reads bright and junk parses read dim. Showing byte runs between executed instructions as explicit data rows is a possible follow-on; it needs realignment logic that did not fit a CSS-class-sized change.

## Repton 2: the comment's test case, finally checked after eleven years

The `prevInstruction` comment has named "Repton 2 @ 2cbb" as a good test case since 2015 (2ce54b1), and the issue's caveat was that game code was never measured. Measured now: `Superior/Repton2.zip` fetched at runtime from the STH mirror, booted headlessly (SHIFT+BREAK, real Video so the loader's vsync waits run), and played into the game via injected keys, with executed starts recorded through the same `debugInstruction` mechanism the debugger uses. Nothing is committed to the repository and no test touches the network.

- During the loader and title music, &2CBB holds `STA $FE40` in the sound-write routine (the same player soundchip.js's `@0x2caa` comment refers to). The heuristic alone answers &2CB9 `ORA #$08`, which is correct; with the map the same answer comes back verified rather than guessed.
- Once the main game loads over the title code, &2CBB holds data and the map self-invalidates: the verified count visibly dropped mid-load as the loader overwrote previously executed memory. The old test case turns out to be an overwritten-RAM case, exactly what the opcode compare is for.
- Accuracy against ground truth (executed starts with a single executed fall-through predecessor, pc outside the window): heuristic alone 529/530 (99.8%) on the loader/title corpus and 1440/1441 (99.9%) in-game; with the map, 100% on both. The one in-game miss is a junk parse `AND ($60,X)` at &087F chosen over the real `RTS` at &0880.

## Tests

Map recording, self-invalidation on overwrite and bank keying; hook install on enable and removal on disable, including the return to the fast execute loop; `prevInstruction` preferring a verified answer, ignoring a stale one, and falling back with nothing verified; the `unexecuted` class present on unexecuted rows and absent on executed ones.

### DIFF
--- a/src/6502.opcodes.js
+++ b/src/6502.opcodes.js
@@ -1105,17 +1105,25 @@ class Disassemble6502 {
      * opcodes are discarded. The highest-scoring run's last instruction wins;
      * any run beats none, since a wrong parse resynchronises within a few
      * instructions and so nearly always shares the true stream's last one.
+     * When an execution map is supplied, a verified executed instruction
+     * ending exactly at the target is returned before any guessing.
      * Good test cases:
      *   Repton 2 @ 2cbb
      *   MOS @ cfc8, and the unrolled STA abs,X screen clear @ cc63
      * also, just starting from the back of ROM and going up...
      */
-    prevInstruction(address, pc) {
+    prevInstruction(address, pc, executionMap = null) {
         const commonInstructions =
             /(RTS|B..|JMP|JSR|LD[AXY]|ST[AXY]|TA[XY]|T[XY]A|AD[DC]|SUB|SBC|CLC|SEC|CMP|EOR|ORA|AND|INC|DEC).*/;
         const uncommonInstructions = /.*,\s*([XY]|X\))$/;
 
         address &= 0xffff;
+        if (executionMap) {
+            for (let length = 1; length <= 3; ++length) {
+                const addr = (address - length) & 0xffff;
+                if (executionMap.isVerified(addr) && (this.disassemble(addr)[1] & 0xffff) === address) return addr;
+            }
+        }
         let bestAddr = address - 1;
         let bestScore = -1;
         for (let startingPoint = address - PrevInstructionWindow; startingPoint !== address; startingPoint++) {
--- a/src/execution-map.js
+++ b/src/execution-map.js
@@ -1,0 +1,49 @@
+"use strict";
+
+const SidewaysStart = 0x8000;
+const SidewaysEnd = 0xc000;
+const SidewaysSize = SidewaysEnd - SidewaysStart;
+const NumBanks = 16;
+
+/**
+ * Records which addresses the CPU has fetched opcodes from. Each entry holds
+ * the opcode byte plus one (zero means never executed), so an entry vouches
+ * for itself: it is trustworthy exactly when the byte in memory still matches,
+ * which covers self-modifying code and overwritten RAM with no write hook.
+ * The sideways region is keyed by ROM bank as each bank has its own code at
+ * the same addresses; Master shadow and private RAM alias that keying, and
+ * the opcode check catches nearly all of it.
+ */
+export class ExecutionMap {
+    constructor(cpu) {
+        this._cpu = cpu;
+        this._main = new Uint16Array(0x10000);
+        this._sideways = new Array(NumBanks).fill(null);
+    }
+
+    record(addr, opcode) {
+        addr &= 0xffff;
+        if (addr >= SidewaysStart && addr < SidewaysEnd) {
+            const bank = this._cpu.romsel & 15;
+            let entries = this._sideways[bank];
+            if (!entries) entries = this._sideways[bank] = new Uint16Array(SidewaysSize);
+            entries[addr - SidewaysStart] = opcode + 1;
+        } else {
+            this._main[addr] = opcode + 1;
+        }
+    }
+
+    /** True when the instruction at addr was executed and its opcode byte is unchanged. */
+    isVerified(addr) {
+        addr &= 0xffff;
+        let entry;
+        if (addr >= SidewaysStart && addr < SidewaysEnd) {
+            const entries = this._sideways[this._cpu.romsel & 15];
+            if (!entries) return false;
+            entry = entries[addr - SidewaysStart];
+        } else {
+            entry = this._main[addr];
+        }
+        return entry !== 0 && entry === this._cpu.peekmem(addr) + 1;
+    }
+}
--- a/src/jsbeeb.css
+++ b/src/jsbeeb.css
@@ -406,6 +406,10 @@ div.modal-body > div:not(:first-child) {
   background-color: #227777;
 }
 
+.dis_elem.unexecuted {
+  opacity: 0.45;
+}
+
 .highlight {
   border: 1px solid #444;
 }
--- a/src/web/debug.js
+++ b/src/web/debug.js
@@ -1,6 +1,7 @@
 "use strict";
 import { hexbyte, hexword, noop, parseAddr } from "../utils.js";
 import { toggle } from "../dom-utils.js";
+import { ExecutionMap } from "../execution-map.js";
 
 const numToShow = 16;
 
@@ -102,6 +103,8 @@ export class Debugger {
         this.disassStack = [];
         this.uservia = this.sysvia = this.crtc = null;
         this.breakpoints = {};
+        this._executionMap = null;
+        this._executionHook = null;
 
         function setupGoto(form, func) {
             const addr = form.querySelector(".goto-addr");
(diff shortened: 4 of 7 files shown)


mattgodbolt:
Wow this seems terrible. I don't think this is right at all:
- half the time we haven't been running code for long with the debugger open. usually we are just looking at code that just stopped WITHOUT any prior debugging
- when it _is_ on it's a slowdown on the hot path
let's talk though

mattgodbolt:
Closed unmerged by decision: the map records only while the debugger is open, but the canonical entry (Ctrl-Home or a breakpoint after running normally) arrives with it empty, so the exact-previous path cannot fire when it is wanted and the dimming would mark freshly-executed code as never run; and while enabled it holds the CPU on the hooked loop. The PR's lasting yield is its validation evidence: the heuristic alone measures 99.8 to 99.9% on real game code after #1017, and the Repton 2 test case from the 2015 comment resolves correctly with no map (its real lesson being overwritten RAM, not heuristic failure).

Issue #1002, opened 2026-08-31
Follow-on design improvements from the post-#971 audits


mattgodbolt:
The same audits that produced #1001 proposed a next round of design work in the vein of #638/#971. Collected here for consideration; none of it is urgent, and deliberately none of it starts until the #1001 bug wave has landed. Each theme is sized by what it touches.

Progress is tracked with the checkboxes below (items struck through were found stale when re-verified against the current tree).

## Theme A: Settings is re-forming into a god object

- [x] `settings.wire({ audioHandler, display, layout, quickSettings, machine, keys, inputs, modals })` (`src/main.js:325`) exists only to dodge construction order, and every apply reaches through the bag; `Settings` even knows that a display-mode change requires `layout.resize()` (`settings.js:126`), a fact only Display and Layout should share. Proposal: `Settings` becomes an EventTarget owning resolution, persistence and the URL, dispatching per-key changes; each module subscribes at its own construction. This dissolves `wire()` and the `targets` bag, the hand-fanned `QuickSettings`/`Config` duplication (`settings.js:106-128`), and `Config`'s double notification channel (three constructor callbacks plus a `settings-changed` event for the same `hide.bs.modal`, `config.js:60-92`). (#1047 guard, #1048: `Settings` is an EventTarget with `set()`/`on()`; Config and QuickSettings are views; consumers subscribe in main.js. Two claims corrected there: the `emulationConfig.keyLayout` write is read on reset and stays; Electron also consumed `settings-changed`.)

## Theme B: loop pause ownership

- [x] "Stop the loop, do a thing, restart it" is written five times by three owners (`modals.js:19-26`, `snapshot-ui.js:111-147`, `rewind-ui.js:58-113`) with independent latches that collide (bug 2 of #1001 is the symptom). One pause-depth mechanism on `EmulationLoop` (a `pause()` returning a resume handle, or `whilePaused(async fn)`) deletes the copies. Whatever minimal fix lands for #1001 item 2, this is the durable shape. (#1045: `pause(reason)` returning an idempotent release; holds nest, a waiting `go()` names its holders)

## Theme C: main.js residue

[...]

PR #1038, opened 2026-09-02
fix: stop the fast-forward and frameSkip frame skips multiplying

mattgodbolt:
Part of #1002.

Two independent frame-skip mechanisms existed. `Display.frameSkip` comes from the `?frameSkip=` URL parameter (and the console benchmarks set it to a million) and drops presents in `onPaint`. `Video.frameSkipCount` is set to 9 by the emulation loop while fast-forwarding or fast-taping and drops whole renders at flyback. With both active they applied in turn, so `?frameSkip=2` during fast-forward showed one frame in 18.

Chosen semantics: the deeper skip wins, and Display owns the composition. The loop now calls `display.setSpeedy(speedy)` instead of poking the video chip. While speedy, the whole skip lives in the video chip at `max(9, frameSkip)`, rounded up to odd so interlaced modes (MODE 7) still alternate fields, and Display presents every frame the chip paints. Outside speedy mode the chip renders every frame and Display presents one in `frameSkip`, exactly as today.

Someone setting only `?frameSkip=` sees identical behaviour (same frames presented, every frame still rendered, which keeps both interlace fields fresh in the framebuffer). Someone only fast-forwarding gets the same one-in-nine as before. Only the combined case changes, from multiplied to the deeper of the two.

Unit tests pin the composition: the skip moving into the chip and back, a deeper configured skip surviving speedy mode, and no double skipping while the chip is skipping.

### DIFF
--- a/src/web/display.js
+++ b/src/web/display.js
@@ -75,7 +75,8 @@ export class Display {
      * configured frameSkip wins, so the two never multiply together.
      */
     setSpeedy(speedy) {
-        this.video.frameSkipCount = speedy ? Math.max(SpeedyFrameSkip, this.frameSkip) | 1 : 0;
+        const skip = Math.max(SpeedyFrameSkip, this.frameSkip);
+        this.video.frameSkipCount = speedy ? (skip % 2 ? skip : skip + 1) : 0;
     }
 
     onPaint(video, minx, miny, maxx, maxy) {
--- a/src/web/display.js
+++ b/src/web/display.js
@@ -2,6 +2,11 @@ import * as canvasLib from "../canvas.js";
 import { FakeVideo, Video } from "../video.js";
 import { toast } from "./toast.js";
 
+// While running fast the state machines still run accurately, but painting is
+// the most expensive part of jsbeeb, so most frames are skipped. Odd, so that
+// interlaced modes (MODE 7) alternate fields across the frames that do paint.
+const SpeedyFrameSkip = 9;
+
 /**
  * The picture: the canvas and its filter, the video chip that paints into a
  * framebuffer of our own, and the animation frame that presents it. A stalled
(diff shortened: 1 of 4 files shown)


src/web/display.js
```diff
@@ -64,10 +69,22 @@ export class Display {
...
     }
 
+    /**
+     * While running fast the skip moves into the video chip, which skips the
+     * render as well as the present; the deeper of the chip's skip and any
+     * configured frameSkip wins, so the two never multiply together.
+     */
+    setSpeedy(speedy) {
+        this.video.frameSkipCount = speedy ? Math.max(SpeedyFrameSkip, this.frameSkip) | 1 : 0;
+    }
```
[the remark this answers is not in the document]
mattgodbolt: Fixed in dd0c12d: the skip is rounded up to odd arithmetically (`skip % 2 ? skip : skip + 1`) rather than with `| 1`, so an absurd `?frameSkip=` stays a large positive count instead of wrapping negative. The pinning test for the rounding (100 to 101) still holds.

src/web/display.js
```diff
@@ -64,10 +69,23 @@ export class Display {
...
+        const skip = Math.max(SpeedyFrameSkip, this.frameSkip);
+        this.video.frameSkipCount = speedy ? (skip % 2 ? skip : skip + 1) : 0;
+    }
+
     onPaint(video, minx, miny, maxx, maxy) {
-        this.frames++;
-        if (this.frames < this.frameSkip) return;
-        this.frames = 0;
+        // The chip paints once from its own constructor, before this.video is set.
+        if (!this.video?.frameSkipCount) {
```
mattgodbolt: why this.?video and not `video` ?

mattgodbolt: No reason; the chip hands itself in as `video`, so the optional chaining and the comment excusing it were both covering for a field that did not need reading. Fixed in 8d437f7: `if (!video.frameSkipCount)`, and the speedy test's fake painter now reports the skip it is running at, as the real chip does.

PR #1046, opened 2026-09-02
test: scale test workers to the machine and use its GPU for e2e

mattgodbolt:
`npm test` times out on a four-core laptop with a browser open. Two causes, both measured there:

- vitest's default of one worker per hardware thread runs seven emulated machines flat out on four cores. Under that, nops ran two to seven times slower than serial and blew its 30s limit; a full run under paging did the same to protection's 10s.
- The display mode smoke test compiles three filters' shaders under software GL. Each compile blocks the page, and while the machine was paging one took over a minute; the action after it was the one that timed out.

What changes:

- `tools/test-workers.js` derives a worker count from hardware threads, the rule `playwright.config.js` already applied ad hoc. Every vitest project now gets one worker per two threads: a worker here is CPU-bound JavaScript (an emulator, or jsdom) and a hyperthread sibling only shares its core. Unit tests got faster, not slower, and it is fewer copies of the process on a machine short of memory.
- vitest is split into `unit`, `integration` and `shader` projects. The integration project owns one timeout, sized as a hang detector, replacing the per-describe overrides that fought it. lint-staged selects the unit project by name, since `--exclude` no longer reaches into projects.
- Playwright points Chrome at the machine's GPU when it has a DRM node, which turns the display mode test from minutes into two seconds here. CI has no GPU and keeps the software renderer it tests today; `E2E_GL=software` or `E2E_GL=hardware` forces either.
- The smoke test waits for the emulator to run again after each mode switch, with the compile budget from #1034 on that wait and on the active-mode check, so the cost lands where it is incurred. Playwright's global timeouts stay as #1034 left them.

Nothing is skipped anywhere. CI's four-thread runner gets two vitest workers instead of three.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -17,7 +17,9 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 - `npm run test:e2e` - Build, then drive the built page in the system Chrome with Playwright and check the things
   unit tests cannot see (main.js constructing everything against the built bundle, Bootstrap layout at real widths, a
   real keystroke, WebGL and WebAudio). Part of `npm test`, and needs Chrome like the shader tests;
-  `BASE_URL=http://localhost:5173/ npx playwright test --project chrome` runs it against a server that is already up
+  `BASE_URL=http://localhost:5173/ npx playwright test --project chrome` runs it against a server that is already up.
+  Chrome uses the machine's GPU when it has one and the software renderer otherwise, which is what CI runs;
+  `E2E_GL=software` or `E2E_GL=hardware` forces one
 - `npm run ci-checks` - Run linting checks for CI
 - `vitest run tests/unit/test-gzip.js` - Run a single test file
 
--- a/playwright.config.js
+++ b/playwright.config.js
@@ -1,4 +1,5 @@
 import { defineConfig } from "@playwright/test";
+import { existsSync } from "node:fs";
 
 import { workersFor } from "./tools/test-workers.js";
 
@@ -9,6 +10,18 @@ const PreviewHost = "127.0.0.1";
 const PreviewPort = 5198;
 const PreviewUrl = `http://${PreviewHost}:${PreviewPort}/`;
 const BaseUrl = process.env.BASE_URL;
+
+// CI's runners have no GPU, so the software renderer is what CI tests. On a
+// machine with one, Chrome is pointed at it instead: a filter's shaders then
+// compile in an instant rather than the minute a paging laptop has taken.
+// E2E_GL=software or E2E_GL=hardware overrides the choice.
+const RenderNode = "/dev/dri/renderD128";
+const GlArgs = {
+    software: ["--use-gl=angle", "--use-angle=swiftshader"],
+    hardware: ["--use-gl=angle", "--use-angle=gl", "--ignore-gpu-blocklist"],
+};
+const Gl = process.env.E2E_GL ?? (process.env.CI || !existsSync(RenderNode) ? "software" : "hardware");
+if (!GlArgs[Gl]) throw new Error(`E2E_GL must be one of ${Object.keys(GlArgs).join(", ")}, not "${Gl}"`);
 
 export default defineConfig({
     testDir: "tests/playwright",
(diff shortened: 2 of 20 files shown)


tests/playwright/smoke.spec.js
```diff
@@ -119,9 +120,12 @@ test("every display mode and sound output on the bar can be picked", async ({ be
...
     for (const mode of modes) {
         // Picking a mode compiles its shaders under software GL, which blocks
-        // the page well past the action default on a slow machine.
-        await page.click(`#display-mode [data-mode="${mode}"]`, { timeout: 30000 });
+        // the page well past the action default on a slow machine; the next
+        // action would otherwise pay for it.
+        const before = await beeb.cycles();
+        await page.click(`#display-mode [data-mode="${mode}"]`, { timeout: ShaderCompileBudgetMs });
         await expect(page.locator("#display-mode .active")).toHaveAttribute("data-mode", mode);
+        await beeb.expectRunningPast(before, ShaderCompileBudgetMs);
```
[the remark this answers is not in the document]
mattgodbolt: Agreed, the traces showed this assertion absorbing seven to nine seconds of compile while the machine paged. It now has the same budget as the click and the wait after it (b982744).

CLAUDE.md
```diff
@@ -17,7 +17,9 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 - `npm run test:e2e` - Build, then drive the built page in the system Chrome with Playwright and check the things
   unit tests cannot see (main.js constructing everything against the built bundle, Bootstrap layout at real widths, a
   real keystroke, WebGL and WebAudio). Part of `npm test`, and needs Chrome like the shader tests;
-  `BASE_URL=http://localhost:5173/ npx playwright test --project chrome` runs it against a server that is already up
+  `BASE_URL=http://localhost:5173/ npx playwright test --project chrome` runs it against a server that is already up.
+  Chrome uses the machine's GPU when it has one and the software renderer otherwise, which is what CI runs;
+  `E2E_GL=software` or `E2E_GL=hardware` forces one
```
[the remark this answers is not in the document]
mattgodbolt: Left as is: none of the entries in this command list end with a period.

playwright.config.js
```diff
@@ -9,31 +11,45 @@ const PreviewPort = 5198;
 const PreviewUrl = `http://${PreviewHost}:${PreviewPort}/`;
 const BaseUrl = process.env.BASE_URL;
 
+// CI's runners have no GPU, so the software renderer is what CI tests. On a
+// machine with one, Chrome is pointed at it instead: a filter's shaders then
+// compile in an instant rather than the minute a paging laptop has taken.
```
mattgodbolt: no need to reference a specific machine in this comment.

playwright.config.js
```diff
@@ -9,31 +11,45 @@ const PreviewPort = 5198;
...
-    // threads keeps small machines serial and big ones parallel.
-    workers: process.env.CI ? 2 : Math.max(1, Math.min(4, Math.floor(os.availableParallelism() / 8))),
+    // Each worker is a whole Chrome running the machine in real time; a 15GB
+    // laptop showed four of them paging each other to death. One worker per
+    // eight hardware threads keeps small machines serial and big ones parallel.
+    workers: process.env.CI ? 2 : workersFor(8, 4),
     reporter: process.env.CI ? [["list"], ["github"], ["html", { open: "never" }]] : "list",
-    timeout: 60000,
+    // A hang detector. The dearest test compiles three filters' shaders under
+    // software GL, which a laptop that is paging has taken over a minute to do.
```
mattgodbolt: again stop mentioning specifics

vite.config.js
```diff
@@ -13,15 +12,25 @@ export default defineConfig({
...
-            "tests/unit/**/test-*.js",
-            "tests/integration/**/*.js",
-            "tests/shader/test-*.js",
-        ],
-        // Playwright's runner owns these; vitest's default globs would take the spec files.
-        exclude: [...configDefaults.exclude, "tests/playwright/**"],
         testTimeout: 15000,
+        // Every worker runs CPU-bound JavaScript (an emulated machine, or jsdom),
+        // so a hyperthread sibling would only share its core: one worker per two
+        // threads. That is also fewer copies of the process on a laptop that is
```
mattgodbolt: enough of the "laptop that's paging"

playwright.config.js
mattgodbolt: Dropped in 13bd376.

playwright.config.js
mattgodbolt: Dropped in 13bd376.

vite.config.js
mattgodbolt: Dropped in 13bd376.

tests/integration/teletext.js
```diff
@@ -72,7 +72,7 @@ async function compare(video, testMachine, expectedName) {
     );
 }
 
-describe("Test Ceefax test page", { timeout: 30000 }, () => {
+describe("Test Ceefax test page", () => {
```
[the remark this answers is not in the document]
mattgodbolt: Pre-existing, but fixed in 3dcee45 since the file is in the diff.

Issue #1072, opened 2026-09-02
Core/web boundary: one construction path for the five front ends (Theme D of #1002)


mattgodbolt:
Theme D of #1002, given its own issue now that themes A to C have landed. Everything below was re-verified against main at 19be970; the line numbers are from that tree.

Five front ends stand up a machine (the web page, Electron, `MachineSession` for the MCP server, `TestMachine` for vitest, the bench tools) and there is no shared construction path between them. The items are ordered so that each one leaves the next easier, and each is its own PR.

- [x] **1. Split `utils.js`.** (#1074) 1282 lines, 33 exports, at least seven unrelated concerns: keymaps and layout detection, archive and zip handling, hex formatting and dumps, data loading (browser and node), binary helpers, the debounce, and Google Analytics. Importing it for `signExtend`, as `6502.js` does, runs `detectKeyboardLayout()` at module scope (`utils.js:569`), which reads `localStorage` and `navigator` and then remaps the exported `keyCodes`; `noteEvent` (`utils.js:973`) reaches for `gtag` and `window.location`. `utils_atom.js:6` carries a duplicate `detectKeyboardLayout`. Proposal: `keymap.js` (BBC and Atom tables plus one layout detector, applied on request rather than at import), `archive.js`, `hex.js`, `loader.js` (with the node branch isolated), `binary.js`, and `analytics.js` in `web/`. Core modules import only what they use, and nothing in `src/` outside `web/` touches a browser global at import time.

- [x] **2. One `buildMachine`.** (#1076: `machineSpec()` and `buildMachine()` in `src/build-machine.js`, `fixUpConfig` gone) `web/machine.js:81-102` and `fake6502.js:19-39` construct the CPU with divergent config objects (the web one passes twelve fields, the fake four), papered over by `fixUpConfig` in `6502.js:571`, which fills in `keyLayout`, `cpuMultiplier`, `userPort`, `printerPort`, `extraRoms` and `debugFlags` for whoever forgot them. Proposal: a single `buildMachine({ model, spec, io })` in `src/`, with a `NullIo` assembled from the existing `Fake*` classes for headless callers; freeze the spec so a missing field is an error, and delete `fixUpConfig`.

[...]

PR #1076, opened 2026-09-03
refactor: one way to build a machine, from a spec that is complete

mattgodbolt:
Item 2 of #1072.

**Before.** `web/machine.js` built its CPU with twelve config fields and `fake6502.js` with four, and `fixUpConfig` in the CPU quietly filled in whatever either had forgotten (`keyLayout`, `cpuMultiplier`, a private `FakeUserPort`, `printerPort`, `extraRoms`, `debugFlags`). Three unit files built CPUs by hand with partial configs of their own. Nothing said what a complete machine was.

**After.** `machineSpec(overrides)` in `src/machine-spec.js` is the one place that does: every field present, defaults in one table, an unknown field an error, an undefined override meaning the default, and the result frozen. `buildMachine({ model, spec, io })` in `src/build-machine.js` constructs the processor for the model, and `nullIo()` is the headless set of peripherals `fake6502` used to assemble by hand. `fake6502` and the web `Machine` both go through it (the web class takes `build` as its injection point, where it took `makeCpu`), and the CPU refuses a config that did not come from `machineSpec`, so a partial one fails at construction rather than somewhere downstream. `fixUpConfig` and `FakeUserPort` are deleted.

**The one runtime write.** `main.js` wrote `emulationConfig.keyLayout` when the layout setting changed, because a reset re-reads it; that cannot be done to a frozen spec, and should not have been done to a config. The processor now keeps `keyLayout` itself and has `setKeyLayout()`, which changes the layout for the keyboard and for every reset after it; `Keyboard.setKeyLayout` calls that instead of reaching into the VIA or PPIA, and the write in `main.js` is gone.

Tests: `test-machine-spec.js` (defaults, unknown fields, undefined overrides, frozenness) and `test-build-machine.js` (the model's CPU class, fitting per spec, the refusal, the layout change reaching the VIA); the three hand-built CPUs in `test-cpu-state`, `test-rom-loading` and `test-snapshot` take a `machineSpec()`; the keyboard tests assert on the processor call. Item 5 of #1072 will move the `isAtom ?` CPU-class pick in `buildMachine` onto `Model` with the others.


src/6502.js
```diff
@@ -645,7 +626,9 @@ export class Cpu6502 extends Base6502 {
         } = {},
     ) {
         super(model, { cycleAccurate });
-        this.config = fixUpConfig(config);
+        if (!Object.isFrozen(config)) throw new Error("A machine's config must come from machineSpec()");
```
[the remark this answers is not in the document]
mattgodbolt: Right, and worse than stated: `Object.isFrozen(undefined)` returns true (non-objects count as frozen), so a missing config would have sailed past the guard and failed later. Fixed in 60c0975 by dropping the frozenness test for a brand: `machineSpec()` stamps its result with a private symbol and exports `isMachineSpec()`, which is what the CPU now checks. A hand-frozen look-alike and `undefined` are both refused with the intended message; tests cover all three cases.

src/machine-spec.js
```diff
@@ -0,0 +1,38 @@
...
+/**
+ * What a machine is fitted with and how it is driven, complete and frozen:
+ * every field is present, an unknown one is an error, and an undefined
+ * override means the default. The CPU takes nothing else.
+ */
+export function machineSpec(overrides = {}) {
+    const unknown = Object.keys(overrides).filter((field) => !(field in SpecDefaults));
+    if (unknown.length) throw new Error(`Unknown machine spec fields: ${unknown.join(", ")}`);
+    const given = Object.fromEntries(Object.entries(overrides).filter(([, value]) => value !== undefined));
+    return Object.freeze({
```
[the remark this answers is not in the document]
mattgodbolt: Fixed in 60c0975: `Object.hasOwn(SpecDefaults, field)` instead of `in`, with a test that `toString` is rejected like any other unknown field.

src/build-machine.js
```diff
@@ -0,0 +1,27 @@
+import { AtomCpu6502, Cpu6502 } from "./6502.js";
+import { Cmos } from "./cmos.js";
+import { FakeDdNoise } from "./ddnoise.js";
+import { FakeMusic5000 } from "./music5000.js";
+import { FakeRelayNoise } from "./relaynoise.js";
+import { FakeSoundChip } from "./soundchip.js";
+import { FakeVideo } from "./video.js";
+
+/** The processor for `model`, fitted as `spec` says, talking to the peripherals in `io`. */
+export function buildMachine({ model, spec, io, cycleAccurate = true }) {
```
mattgodbolt, line 57: this is a lot of import and whatnot for a two line funciton. is this file justified compared to having this in machine spec? or sometwhere else?

src/build-machine.js
```diff
@@ -0,0 +1,27 @@
...
+    const CpuClass = model.isAtom ? AtomCpu6502 : Cpu6502;
+    return new CpuClass(model, { ...io, config: spec, cycleAccurate });
+}
+
+/** Peripherals that go nowhere, for a machine run headless. */
+export function nullIo({ video = new FakeVideo(), soundChip = new FakeSoundChip() } = {}) {
```
mattgodbolt, line 64: and nullio takes up more of "build machine.js" than any buildMachine... maybe this file is extraneous? maybe this too in machine spec? or somewhere else?

src/machine-spec.js
```diff
@@ -0,0 +1,46 @@
+const NullUserPort = {
```
mattgodbolt: other null stuff defined here, so maybe the right home for nullio?

mattgodbolt (commented):
combine build-machine and machine-spec or something, the two files are tin yand are more import than functionality

src/build-machine.js
mattgodbolt: Done in 24d0b0d: one file, `src/build-machine.js`, holding the spec (with the null user port), `nullIo` and `buildMachine`; `machine-spec.js` is gone and the two test files are one. The CPU's brand check imports from it, which makes `6502.js` and `build-machine.js` refer to each other; the build raises no circular-dependency warning and nothing is used at module evaluation time, so it is a static cycle only.

src/build-machine.js
mattgodbolt: Done in 24d0b0d: one file, `src/build-machine.js`, holding the spec (with the null user port), `nullIo` and `buildMachine`; `machine-spec.js` is gone and the two test files are one. The CPU's brand check imports from it, which makes `6502.js` and `build-machine.js` refer to each other; the build raises no circular-dependency warning and nothing is used at module evaluation time, so it is a static cycle only.

src/machine-spec.js
mattgodbolt: Done in 24d0b0d: one file, `src/build-machine.js`, holding the spec (with the null user port), `nullIo` and `buildMachine`; `machine-spec.js` is gone and the two test files are one. The CPU's brand check imports from it, which makes `6502.js` and `build-machine.js` refer to each other; the build raises no circular-dependency warning and nothing is used at module evaluation time, so it is a static cycle only.

mattgodbolt:
Combined as asked, in 24d0b0d: `src/build-machine.js` is now the one module (spec, null peripherals, builder), `machine-spec.js` is deleted, and the two test files are one. Replies on the three threads say the same.

mattgodbolt:
Cycle removed in 26eb845: the spec check moved from the CPU constructor into `buildMachine`, so `6502.js` no longer imports `build-machine.js`. The three unit files that still built a `Cpu6502` by hand now go through `buildMachine` like everything else.

PR #1079, opened 2026-09-06
refactor: the model knows the shape of its machine

mattgodbolt:
Item 5 of #1072.

**Before.** What a model implies about its machine was re-derived with `isAtom ?` wherever it was needed: the CPU class in `buildMachine`, the WRCHV address in `test-machine.js`, the keyboard chip and shift key in `keyboard.js` and again in `test-machine.js`, the key table in `main.js`, the string-to-keys in `autoboot.js`, the paste cadence in `keyboard.js`, and the wait-for-input address split between `basicIdleAddr` (B or Master) and an Atom-only branch in `TestMachine.runUntilInput`. `basicIdleAddr` showed the right pattern (a fact of the model, in one place) and this extends it.

**After.** `Model` carries `Cpu`, `keys`, `stringToKeys`, `wrchvAddress`, `idleAddress`, `pasteKeyDelayMs` and `keyboardOf(processor)`, all set in its constructor, which is now the one place the two machine families part. The dispatch sites read them; `basicIdleAddr` is gone (the model's `idleAddress` covers the Atom too, so `runUntilInput` has one path); the imports of the Atom key map leave `main.js`, `keyboard.js` and `autoboot.js`. `models.js` gains imports of the CPU classes and the two key maps; nothing on those paths imports `models.js`, so there is no cycle, and the build agrees. With the CPU class on the model, `buildMachine` had nothing left to do but the spec check, so it and the check are gone (review): callers construct `new model.Cpu(model, { ...io, config: machineSpec(...) })`, and the module is `machine-spec.js`, holding `machineSpec` and `nullIo`.

**Left as `isAtom` checks, on purpose:** the VDU decoder's choice of sequence set (an option to a decoder that has no model), the audio worklet's sound chip choice (a class cannot cross the worklet boundary, so a flag must), the session's instrumented sound chip (test instrumentation, not machine shape), and the front panel's tape control (a UI toggle). The tape format code's `isAtom` is about the tape encoding, not the machine.

Tests: `test-models.js` checks each fact for a B, a Master and an Atom; the keyboard, keyboard-setup and autoboot tests use real models where the code now asks the model (their `{ isAtom }` fakes could not answer); the autoboot tape test asserts on the keys sent rather than spying on a function the model now holds by reference; the idle-address test moves from the BASIC loader to the model.

### DIFF
--- a/src/basic-loader.js
+++ b/src/basic-loader.js
@@ -1,5 +1,3 @@
-/** Where the OS sits waiting for a keypress, per machine: the place to interrupt with a program. */
-
 /**
  * Pokes a tokenised BASIC program into memory at PAGE, and sets TOP and
  * VARTOP after it, exactly as if it had just been typed.
--- a/src/basic-loader.js
+++ b/src/basic-loader.js
@@ -1,7 +1,4 @@
 /** Where the OS sits waiting for a keypress, per machine: the place to interrupt with a program. */
-export function basicIdleAddr(model) {
-    return model.isMaster ? 0xe7e6 : 0xe581;
-}
 
 /**
  * Pokes a tokenised BASIC program into memory at PAGE, and sets TOP and
--- a/src/build-machine.js
+++ b/src/build-machine.js
@@ -1,4 +1,3 @@
-import { AtomCpu6502, Cpu6502 } from "./6502.js";
 import { Cmos } from "./cmos.js";
 import { FakeDdNoise } from "./ddnoise.js";
 import { FakeMusic5000 } from "./music5000.js";
@@ -56,8 +55,7 @@ export function isMachineSpec(config) {
 /** The processor for `model`, fitted as `spec` says, talking to the peripherals in `io`. */
 export function buildMachine({ model, spec, io, cycleAccurate = true }) {
     if (!isMachineSpec(spec)) throw new Error("A machine's config must come from machineSpec()");
-    const CpuClass = model.isAtom ? AtomCpu6502 : Cpu6502;
-    return new CpuClass(model, { ...io, config: spec, cycleAccurate });
+    return new model.Cpu(model, { ...io, config: spec, cycleAccurate });
 }
 
 /** Peripherals that go nowhere, for a machine run headless. */
--- a/src/web/autoboot.js
+++ b/src/web/autoboot.js
@@ -66,7 +66,7 @@ export class Autoboot {
         const tokenised = await t.tokenise(prog);
 
         const { processor } = this;
-        const idleAddr = processor.model.idleAddress;
+        const idleAddr = this.model.idleAddress;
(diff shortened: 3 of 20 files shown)


src/web/autoboot.js
```diff
@@ -68,7 +66,7 @@ export class Autoboot {
         const tokenised = await t.tokenise(prog);
 
         const { processor } = this;
-        const idleAddr = basicIdleAddr(processor.model);
+        const idleAddr = processor.model.idleAddress;
```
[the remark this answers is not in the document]
mattgodbolt: Done in 4940374: `insertBasic` reads `this.model.idleAddress`, so Autoboot follows the model it was constructed for everywhere. The mix predates this PR (it used `basicIdleAddr(processor.model)`), but you are right that the refactor is the moment to settle it.

tests/unit/test-autoboot.js
```diff
@@ -83,7 +84,7 @@ describe("Autoboot", () => {
         });
 
         it("hooks the Master's idle loop on a Master", async () => {
-            processor.model.isMaster = true;
+            processor.model = findModel("Master");
             await make().insertBasic(Promise.resolve("10 END"), false);
```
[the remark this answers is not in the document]
mattgodbolt: Done in 4940374: the Master case constructs its own Autoboot with the Master model rather than swapping the processor's model under a B-built one, so the two cannot diverge.

src/basic-loader.js
```diff
@@ -1,7 +1,4 @@
 /** Where the OS sits waiting for a keypress, per machine: the place to interrupt with a program. */
```
mattgodbolt: orphaned comment

src/build-machine.js
```diff
@@ -56,8 +55,7 @@ export function isMachineSpec(config) {
 /** The processor for `model`, fitted as `spec` says, talking to the peripherals in `io`. */
 export function buildMachine({ model, spec, io, cycleAccurate = true }) {
```
mattgodbolt: this function is even more ridiculous than in the previous PR. Remind me its purpose again?

src/basic-loader.js
mattgodbolt: Gone in 0cf232c; the function left and its doc stayed.

src/build-machine.js
mattgodbolt: Its purpose was to be the one construction path when it also chose the CPU class from the model. This PR moved that choice onto `model.Cpu`, which left buildMachine as two lines whose only job was the spec check, and the check was never the point. So in a0cab83 it is gone: callers write `new model.Cpu(model, { ...io, config: machineSpec(...) })` (fake6502, the web Machine's default builder, three test files), the brand check goes with it, and the file is renamed `machine-spec.js` for what remains, `machineSpec` and `nullIo`.

Issue #824, opened 2026-08-11
One disc picker: both drives, all sources, and a place to put the disc


mattgodbolt:
There are six ways to get a disc into jsbeeb and they agree on almost nothing. Only one of them can reach drive 1, and that one is a URL parameter. Nothing anywhere tells you what is currently in either drive.

This is a design sketch for replacing all of it with a single picker. It is deliberately opinionated, and where I am unsure I have said so under "Open questions" rather than hedging.

## The problem

#496 has been open since 2023 and states it well: no way to know what is loaded, an awkward menu, no way to unload, and a partition between "examples or local", "from STH", "from Google Drive" and now "from the HFE archive" that users have to guess between. #11 ("UI support for mounting two discs") has been open longer still. In #385 a user describes mounting a disc "in drive 1 using the URL hack", which is the only way there is.

Meanwhile the source list keeps growing. The HFE archive (#800) added a fourth modal in the same shape as the other three, and its manifest carries better metadata than anything else we have, which no part of the UI outside that one modal can see.

## What exists today

Read against `origin/main` at `602be9e`.

### Every route a disc takes into a drive

| Entry point | Code | Drives it can reach | 40/80 respected | Progress shown | On failure | URL updated |
|---|---|---|---|---|---|---|
| Discs, From STH archive | `discSthClick` | 0 only | yes | loading modal | toast | yes, `sth:` |
| Discs, From HFE archive | `hfeClick` | 0 only | yes, since #814 | loading modal | toast | yes, `hfe:` |
| Discs, From examples or local (built in list) | `discList` click handler | 0 only | yes | none | nothing: unhandled rejection, silent | yes, bare name |
| Discs, From examples or local (file input) | `#disc_load` change | 0 only | yes | none | nothing, silent | deletes `disc1` |
| Drag and drop onto the paste box | `pastetext` drop | 0 only | yes | none | nothing, silent | deletes `disc1` |
| Discs, From Google Drive (list) | `googleDriveEl` show handler | 0 only | yes | loading modal | toast | yes, `gd:` |
| Discs, From Google Drive (create) | `#google-drive form` submit | 0 only | drive switch yes, layout not passed to `create` | loading modal | toast | yes, `gd:` |
| `?disc=` / `?disc1=` | `startImageLoad` | 0 | yes | none | toast, boot continues (#808) | n/a |
| `?disc2=` | `startImageLoad` | 1 | yes | none | toast | n/a |
[...]

PR #1089, opened 2026-09-07
fix: media loads go through the drive funnels, and the URL names only what is in the drive

mattgodbolt:
The plumbing fixes ahead of the #824 media window, each a bug on main today and small enough to read alone. Nothing here changes what the page looks like.

- **The desktop app's File menu loads go through the funnels.** Electron's Load disc 0/1 called `fdc.loadDisc` directly, skipping the drive's 40/80 switch, the unsaved-writes notice and the URL, and a file that would not load failed silently. Tape loads likewise bypassed `setProcessorTape` and never named the tape in the URL. Both now report failures like every web route.
- **A disc is named in the URL only once it is in the drive.** The STH, HFE, Google Drive and built-in routes wrote `disc1` (or `tape`) before fetching, so a failed load left the URL naming a disc the drive did not hold.
- **A disc created on Google Drive is laid out for the drive it goes into.** `create()` dropped the layout, so a drive fixed at 80 track still had a 40 track catalogue sniffed and double stepped.
- **The drives and the deck say what they hold, and can be emptied.** `Drives` raises `disc-changed` from `putDiscIn` and the new `eject`; `MediaLoader` raises `tape-changed` and gains `ejectDisc` and `ejectTape`, which also clear the slot's URL parameter. Nothing calls eject yet.
- **`discToDownload` takes the drive.** Both menu links still ask for drive 0.
- **`dfsCatalogue(disc, side)` reads the DFS title and cycle number** off the two catalogue sectors of track 0, which is what was on the sticker.

The Electron hooks had no unit test; they have one now.

### DIFF
--- a/src/web/media-loader.js
+++ b/src/web/media-loader.js
@@ -144,10 +144,10 @@ export class MediaLoader extends EventTarget {
             elem.querySelector(".description").textContent = image.desc;
             elem.addEventListener("click", async () => {
                 noteEvent("images", "click", image.file);
-                this.setDisc1Image(image.file);
                 modals.hide("discs");
                 try {
-                    drives.putDiscIn(0, await this.loadDiscImage(this.params.disc1, drives.layoutForDrive(0)));
+                    drives.putDiscIn(0, await this.loadDiscImage(image.file, drives.layoutForDrive(0)));
+                    this.setDisc1Image(image.file);
                 } catch (error) {
                     reportLoadFailure(`${image.name} (${image.file})`, error);
                 }
--- a/src/web/media-loader.js
+++ b/src/web/media-loader.js
@@ -165,8 +165,21 @@ export class MediaLoader extends EventTarget {
         else this.resolver.addSource(schema, fetcher);
     }
 
+    /** Puts a tape in the deck, or empties it; raises "tape-changed" with what the deck now holds. */
     setProcessorTape(tape) {
         this.processor.tapeInterface.setTape(tape);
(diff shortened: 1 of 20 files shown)


src/web/media-loader.js
```diff
@@ -165,8 +165,21 @@ export class MediaLoader extends EventTarget {
...
 
+    /** Puts a tape in the deck, or empties it; raises "tape-changed" with what the deck now holds. */
     setProcessorTape(tape) {
         this.processor.tapeInterface.setTape(tape);
+        this.dispatchEvent(new CustomEvent("tape-changed", { detail: { tape } }));
+    }
+
+    ejectDisc(driveIndex) {
+        this.drives.eject(driveIndex);
+        if (driveIndex === 0) this.setDisc1Image(undefined);
```
mattgodbolt: I'm not a fan of setDisc1Image() etc - is there a reason why we don't have `setDiscImage(drive, image)` - seems ... odd to name the drive that way in the media loader. I realise that's out of scope here but thoughts?

mattgodbolt: Agreed: the names came from the URL parameters, from before drive 1 was reachable from anything but the URL. Done in 849f4c1: `setDiscImage(driveIndex, name)` writes `disc1` or `disc2` (and clears the bare `disc` for drive 0) and the two call sites that branched on the drive collapse to one line. The URL parameters and the `media-changed` detail keep their `disc1`/`disc2` keys, since the Electron settings store reads those.

Issue #1059, opened 2026-09-02
WD1770 loose ends: C2 sync, marker byte range, state timing, 1772 and Opus variants


mattgodbolt:
The TODO comments left in `src/wd-fdc.js` by the beebjit port, collected so someone working on the 1770 has them in one place. Three are accuracy notes about the bit-level state machine; two mark controller variants we have flags for but do not model.

Accuracy:

- `src/wd-fdc.js:968`: sync to c2 (5224). MFM mark detection recognises the 0xA1 sync (bit pattern 0x4489) but not the 0xC2 sync (0x5224) that precedes an index address mark. The comment notes an earlier naive attempt fired in the middle of sector data, so it needs studying properly.
- `src/wd-fdc.js:983`: see http://info-coach.fr/atari/documents/_mydoc/WD1772-JLG.pdf. This suggests that a wider range of byte values will function as markers. It may also differ FM vs. MFM. We accept only 0xF8, 0xFB and 0xFE with a 0xC7 clock as FM markers.
- `src/wd-fdc.js:1224`: it's likely that timing is generally off for most states, i.e. the 1770 takes various numbers of internal clock cycles before it delivers the CRC error, before it goes not busy, etc. Noted at the end of read address, where we go straight to `State.done`.

Variants:

- `src/wd-fdc.js:150`: `_is1772 = false`, if we ever support Master Compact. The 1772 differs in step rates.
- `src/wd-fdc.js:151`: `_isOpus = false`, if we ever support Opus. The Opus board wires INTRQ differently, which `_updateNmi` already allows for.

The code comments now reference this issue.

PR #1114, opened 2026-09-09
fix: read track on the 1770 streams the track instead of failing as a sector read

mattgodbolt:
Found while working on #1059. On main, `State.inReadTrack` had the same value as `State.inData`, and the pulse handler's switch listed `State.readTrack`, a name that does not exist in the enum. So a Read Track command on a 1770 machine (Master, B+) went through the sector data path: it stopped after one sector's worth of bytes and reported a CRC error, delivering nothing useful. Any disc utility or copy protection that reads a whole track on a Master got garbage.

The fix gives `inReadTrack` its own value (appended after `done`, so no other saved state number changes meaning) and corrects the case label. The test formats one FM sector on a track, issues Read Track, and checks the stream covers the revolution, contains the ID and data marks with their headers, and reports no CRC error; it fails on main.

The same two-line change is also on the #1059 branch, which depends on it; the branches merge cleanly in either order.


src/wd-fdc.js
```diff
@@ -108,7 +108,7 @@ const State = Object.freeze({
     inId: 6,
     searchData: 7,
     inData: 8,
-    inReadTrack: 8,
+    inReadTrack: 19,
```
mattgodbolt: please put this in the correct order. should it be 18, and done: 19?

mattgodbolt: Done in aeb716c: inReadTrack is 9 after inData, and everything after shifts up one, done ending at 19. One consequence worth knowing: the WD FDC snapshot saves this value as a bare number, so a snapshot taken mid-command with a state of 9 or above (write sector, write track, checkMulti, done) restores as the state one before it. Mid-command snapshots are rare and the old value 8 was already ambiguous, so I have not bumped the format version. The #1059 branch is getting the identical enum block so the two merge cleanly.

Issue #1061, opened 2026-09-02
Disc code tidy-ups: track builder and reader, DiscConfig, debug logging, NoiseAware dedupe, FM pulse bit position


mattgodbolt:
Small notes left in the disc code that are each too minor for an issue on their own but together make a reasonable tidy-up pass over `src/disc.js` and its neighbours. None of them changes behaviour except the last, which is a question about correctness of a helper.

- `src/disc.js:72`: consider remembering isMfm if nothing else needs to know; could then break this into MFM and FM builder. `TrackBuilder.appendCrc` takes an `isMfm` argument on every call even though a track is one or the other.
- `src/disc.js:307`: put in reader. `pulsesPerByte = this.isMfm ? 16 : 32` is computed inside `Sector.read` rather than living with the pulse reader.
- `src/disc.js:539`: is this even useful? `DiscConfig` carries `logProtection`, `logIffyPulses`, `expandTo80`, `isQuantizeFm`, `isSkipOddTracks`, `isSkipUpperSide`; check which are read anywhere.
- `src/disc.js:1056`: a debug log flag for this. A commented-out `console.log` of every pulse write in `writePulses`.
- `src/disc.js:1227`: add fingerprintings, catalog etcetc. At the end of the disc summary logger, which currently reports sectors and CRC errors only.
- `src/wd-fdc.js:1438`: consider deduplicating with the IntelFdc equivalent. `NoiseAwareWdFdc` and `NoiseAwareIntelFdc` carry near-identical spin and seek noise wrappers.
- `tests/unit/test-disc.js:88`: either fix these or understand why beebjit doesn't use same bit posn for bits. The `_2usPulsesToFm` tests shift their input by a `deliberateFudge` of one bit to pass, so either our decoder or the test's expectation is one bit off relative to beebjit.

The code comments now reference this issue.

PR #1115, opened 2026-09-09
fix: FM pulses are written in the phase the decoder samples, so fuzzy bits read as fuzzy

mattgodbolt:
Found while working on #1061, whose test for the FM pulse conversion had to shift its input by a "deliberate fudge" of one bit to pass.

An FM bit cell is four 2 us slots. `IbmDiscFormat.fmTo2usPulses` wrote the clock pulse in the second slot and the data pulse in the fourth, while `_2usPulsesToFm` samples the first and third and treats the other two as off-clock ("iffy") pulses. Every reader in the codebase locks on to the decoder's phase for itself (the sliding mark detectors in `disc.js` and the 1770, and the 8271's bit-serial state machine), so the mismatch did nothing on ordinary tracks. beebjit has the same pair of functions with the same mismatch.

It was not harmless for the 1770's fuzzy-bit path. When an FM byte arrives with off-clock pulses (MFM data read as FM, the classic weak-bit protection), `_bitReceived` substitutes a decode of the drive's quasi-random pulses, which the comment there says should give a non-stable read. Those pulses come from the same encoder, so under the old phase they decoded as data 0x00 with the iffy flag set, every time: the "fuzzy" bits were a stable zero.

The encoder now writes the phase the decoder samples, so the two are exact inverses; the unit tests assert that directly and the fudge is gone. A new test checks the drive's quasi-random pulses decode cleanly and change over time, and fails on main.

Two knock-on effects, both corrections rather than regressions:

- Fuzzy bits on the 1770 (Master, B+) now read as varying bytes. The 8271 has no such path, so the copy-protection integration tests (which run on the B with DFS 1.2) do not depend on the old stable zero; the full integration suite passes.
- `Track.findSectorIds` now reports FM `idPosBitOffset` and `dataPosBitOffset` on the word boundary rather than one bit past it, so the disc surface view's `fill` no longer spills one word into the following gap for FM sectors. MFM offsets were already exact.

The same encoder change is also on the #1061 branch, which depends on it; the branches merge cleanly in either order.

### DIFF
--- a/src/disc.js
+++ b/src/disc.js
@@ -1348,6 +1348,7 @@ export class IbmDiscFormat {
             clocks = (clocks << 1) & 0xff;
             data = (data << 1) & 0xff;
         }
+        // Bit 7's clock lands in bit 31, so back to unsigned.
         return ret >>> 0;
     }
 
--- a/src/disc.js
+++ b/src/disc.js
@@ -1343,12 +1343,12 @@ export class IbmDiscFormat {
         let ret = 0;
         for (let i = 0; i < 8; ++i) {
             ret <<= 4;
-            if (clocks & 0x80) ret |= 0x04;
-            if (data & 0x80) ret |= 0x01;
+            if (clocks & 0x80) ret |= 0x08;
+            if (data & 0x80) ret |= 0x02;
             clocks = (clocks << 1) & 0xff;
             data = (data << 1) & 0xff;
         }
-        return ret;
+        return ret >>> 0;
     }
 
     static _2usPulsesToFm(pulses) {
--- a/tests/unit/test-disc-drive.js
+++ b/tests/unit/test-disc-drive.js
@@ -72,6 +72,18 @@ describe("Disc drive tests", function () {
         scheduler.polltime(1000000);
         expect(called).toBe(true);
     });
+    it("makes quasi random pulses the FM decoder reads as clean data that changes over time", () => {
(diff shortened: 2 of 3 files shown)


src/disc.js
```diff
@@ -1343,12 +1343,12 @@ export class IbmDiscFormat {
...
             ret <<= 4;
-            if (clocks & 0x80) ret |= 0x04;
-            if (data & 0x80) ret |= 0x01;
+            if (clocks & 0x80) ret |= 0x08;
+            if (data & 0x80) ret |= 0x02;
             clocks = (clocks << 1) & 0xff;
             data = (data << 1) & 0xff;
         }
-        return ret;
+        return ret >>> 0;
```
mattgodbolt, line 1352: what is this?

mattgodbolt: It reinterprets the word as unsigned. JavaScript bitwise operators work on signed 32-bit values, and with the clock now in the first slot of each nibble, bit 7's clock lands in bit 31, so `ret` would come back negative for any byte whose top clock is set (0x88888888 for a sync byte). The old encoder's top slot was bit 30 so it never needed it. `Uint32Array` storage and every masking consumer are unaffected either way; only comparisons and the tests see the sign. `& 0xffffffff` would not help (still a signed result) and `& 0x7fffffff` would drop the clock. Noted in the code in 33bb322.

Issue #1134, opened 2026-09-12
Emulated frames are lost between flyback and the animation frame that presents them


mattgodbolt:
Since #891 (the emulator advancing on a 10 ms timer instead of `requestAnimationFrame`), a completed emulated frame is sometimes overwritten by the next one before the animation frame that would have presented it fires. On a 60 Hz display this loses about one frame in ten under load and roughly one in sixty on a quiet machine, visible as a two-frame positional jump in anything that moves every frame. The old loop did not lose frames at 60 Hz unless the browser itself skipped a vsync.

## Repro

Cycle the border through red, yellow, green, cyan, one colour per vsync:

```
10 MODE 2
20 DIM C%(3):C%(0)=1:C%(1)=3:C%(2)=2:C%(3)=6
30 I%=0
40 REPEAT
50 *FX 19
60 VDU 19,0,C%(I%),0,0,0
70 I%=(I%+1) MOD 4
80 UNTIL FALSE
```

[Run it on bbc.xania.org](https://bbc.xania.org/?embedBasic=10%20MODE%202%0A20%20DIM%20C%25(3)%3AC%25(0)%3D1%3AC%25(1)%3D3%3AC%25(2)%3D2%3AC%25(3)%3D6%0A30%20I%25%3D0%0A40%20REPEAT%0A50%20*FX%2019%0A60%20VDU%2019%2C0%2CC%25(I%25)%2C0%2C0%2C0%0A70%20I%25%3D(I%25%2B1)%20MOD%204%0A80%20UNTIL%20FALSE%0A). A 240 fps phone recording of a 60 Hz screen shows the expected 50-into-60 repeats, and every so often a colour missing from the sequence (red, yellow, cyan). The same program on a build before #891 shows every colour in order.

To count it without a camera, from the console: wrap `window.video.paint_ext` (one call per emulated flyback) and `WebGLRenderingContext.prototype.drawArrays` (one per presented frame); flybacks minus draws is the number lost. Wrapping `scheduler.postTask` gives the real tick period.

## Measured

1500 emulated frames per run, Chrome 151. The 60 Hz runs are `--headless=new`, whose synthetic 60 Hz vsync goes through the same scheduler; the 30 Hz runs are a real 4K-over-HDMI display.

| Build | 60 Hz | 30 Hz |
|---|---|---|
| Before #891 (rAF-driven) | 43 lost, all from the browser skipping a vsync under software GL | 601 lost |
[...]

PR #1135, opened 2026-09-13
fix: completed frames queue for an animation frame each instead of the newest overwriting the rest

mattgodbolt:
Part of the fix for #1134; the third of three independent changes (with #1136 and #1137). This one carries the most latency risk and does the least on a healthy machine, so read the numbers before judging it.

## What changed

`Display` keeps a queue of up to two finished frames instead of one slot the newest flyback overwrote. Each animation frame presents the oldest and asks for another animation frame if any remain. On overflow the oldest is dropped. A waiting frame older than one frame period gives way to the newer one behind it, so a display slower than the machine, a stall or a hold collapses to newest-wins instead of showing stale frames one per refresh. A present with the frame time it was last given draws nothing, so the loop's vsync present in #1137 and the display's own animation frame take at most one frame per refresh between them. Frame buffers are pooled.

## Measured

1500 emulated frames per run. "Quiet 60 Hz" is `--headless=new --use-angle=vulkan` (ticks at 10.2 ms, steady vsync). "Loaded 60 Hz" is software-GL headless on a busy box (ticks at 14 ms median, rAF p99 33 ms). Real 30 Hz is a 4K panel over HDMI.

| | quiet 60 Hz lost | loaded 60 Hz lost | real 30 Hz lost |
|---|---|---|---|
| `main` | 34, 21 | 103, 166 | 600 |
| this PR | 24, 10 | 93 | 600 |
| #1136 + #1137, no queue | 0, 0 | 114 | 601 |
| all three | 1, 1 | 94, 109 | 601 |

Shown-frame age (flyback to GL upload) under load: p50 4.0 ms, p90 15.6, p99 20.0 against `main`'s 3.0, 8.3, 16.4. A simulated 30 Hz run (every other animation frame dropped) put the median age at 12.6 ms against `main`'s 7.4; the plain two-deep queue without the staleness bound was 23.0.

## Trade-offs, plainly

- **Latency.** Showing both frames of a bunched pair means the second is shown a refresh later than newest-wins would, and after a bunch the queue stays a frame behind until the spare vsync per 100 ms absorbs it, which is the p90 tail above. The staleness bound caps that at about a frame period and removes the cost on slow displays, but it is a real cost at 60 Hz and it is the reason for the extra loss this PR still shows under load: a bunched frame older than 20 ms is dropped again.
- **It is not what fixes the healthy-machine case.** #1136 and #1137 take a quiet 60 Hz machine from 34 lost to 0 without a queue. The queue's only distinct contribution is on a machine where animation frames come late, and there the bound gives most of it back. All three together are no better than the two loop changes at quiet 60 Hz and about the same loaded.
- **After a stall** the bound skips straight to the newest frame, which is `main`'s behaviour; the plain queue would have shown two stale frames first.
- **An extra copy per present** (queued rows into the canvas upload buffer, about 0.3 ms for a full frame) and two extra 2.5 MB framebuffers. Uploading straight from the queued buffer would remove the copy at the cost of teaching both canvases to take a source buffer.

## Testing

`tests/unit/test-display.js` covers the two-frames-one-each ordering, overflow dropping the oldest, the staleness bound and a lone old frame, one draw per animation frame, buffer reuse, and the interlace metadata travelling with the frame. Not felt on a real 60 Hz display.


mattgodbolt:
This _may_ be unnecessary, pending feedback from kenton

mattgodbolt:
Parked while Kenton retests on the merged site. State for whoever picks this up:

- **Where the loop changes got to.** With #1136 and #1137 live, a real 60 Hz laptop with Starquake running lost 13 frames in 6534 (about one in 500, one every ten seconds), down from roughly one in fifty before. The residual is a vsync callback arriving late enough that a timer tick has already produced the next flyback, plus the occasional skipped vsync. That residual is the only thing left for this PR to address.
- **The staleness bound as written does not catch a skipped vsync.** A skipped vsync opens a 33 ms gap; by the next callback the older of the two waiting frames is past the 20 ms bound and is dropped, which is the same loss newest-wins has. To take the residual to zero the bound would have to be at least one refresh plus a frame period (about 37 ms at 60 Hz), or go entirely, at the latency tail measured in the body (plain two-deep queue: p90 age 20 ms against main's 8 at 60 Hz, a full refresh behind at 30 Hz).
- **It will not merge as is.** #1137 changed `src/web/display.js` after this branch was cut: the pending-present flag is now `presented`, and `presentScheduled` is cleared in the animation-frame callback rather than in `present`. When rebasing, take this branch's `onPaint` and `present` wholesale and drop `presented`, since the queue's emptiness is that state; keep `present(frameTime)` and the frame-time guard, which is what stops the loop's vsync present and the display's own animation frame each taking a frame in the same refresh (without it the combined build drew twice per refresh and looked better than it was). The throwaway merge that was measured did exactly that and passed both test files.
- **How to measure.** Flybacks against GL draws in a real Chrome: wrap `window.video.paint_ext` and `WebGLRenderingContext.prototype.drawArrays`, paints minus draws is frames lost; count only while `document.visibilityState` is visible, since a disc motor keeps the machine running in a hidden tab with nothing to present to. For a quiet 60 Hz machine without one to hand, `google-chrome --headless=new --use-angle=vulkan` gives 10.2 ms ticks and a steady vsync; plain `--headless=new` is software GL and behaves like a loaded machine. The BASIC repro in #1134 or Kenton's disc both work as the source.

mattgodbolt:
Closing unmerged: with #1136 and #1137 live, Kenton's 240 fps recording of his diagnostic disc shows 680 consecutive transitions with nothing skipped, so the residual this queue would cover is theoretical for now. The parking note above says what it would take to pick up; #1134 links here.

Issue #885, opened 2026-08-26
Audio: emulation only advances inside requestAnimationFrame, so any display presentation stall is an audio gap


mattgodbolt:
The emulator only advances inside the `requestAnimationFrame` callback (`draw()` in `src/main.js`), and the sound chip only produces samples while the emulator advances. So the audio worklet is fed exactly as regularly as the browser presents frames, and any stall in *presentation* becomes a hole in the audio of the same length, however healthy the machine is.

Measured on a Chrome Performance trace of an otherwise idle machine at the BASIC prompt playing `SOUND 1,-15,200,200`, at the moment of an audible blip:

- Our frame completed normally (`draw` 9.5 ms, commit 1.3 ms).
- The compositor then sat in `ReceiveCompositorFrameToStartDraw` for 76 ms waiting to get that frame onto the display. While a frame is pending, Chrome withholds the next BeginMainFrame (four `DroppedFrame` markers), so no rAF fired.
- The renderer main thread was idle for those 78 ms: no GC, no long task, no script of ours.
- rAF then fired, `draw` caught up 88 ms of emulation in one burst, and the worklet had already run dry.

Before the stall every frame's submit-to-presentation was a steady ~24 ms; after it, ~41 ms with `Swap` blocking a full ~16 ms per frame for the rest of the trace, so something on the display side (system compositor, driver, mode) hiccupped and settled into a slower state. That is outside our control, and on that machine it recurs roughly every 10 s. Deepening the worklet's buffer to cover it costs directly in audible lag (40 ms already shows as a visible sound/picture offset), so a buffer is the wrong tool.

Proposal: decouple the emulation step from rAF. Run `draw()`'s emulation on a timer (the `useTimeout` path already exists for tape and fast-forward) or a rAF with a timeout fallback, and let the animation frame only present whatever the video emulation last painted. Then a presentation stall costs a repeated video frame instead of an audio gap, at no added latency. Care is needed not to double-run when both fire, and the video paint still needs to land on a real frame.

Bigger alternative, for later: synthesise the chip inside the worklet from timestamped register writes (the TODO already in `audio-renderer.js`), so a main-thread stall just holds the current chip state rather than starving the output at all.

Context: this is the mechanism behind the underruns in #884; #865 covers how an underrun sounds once it happens.

mattgodbolt:
[...]

PR #1137, opened 2026-09-13
fix: run the machine up to each vsync and present in the same callback

mattgodbolt:
Part of the fix for #1134; the second of three independent changes, stacked on #1136 (with #1135 the third). Read the trade-offs before judging it.

## What changed

While running, the loop also takes an animation frame on every vsync. In that callback it runs the machine only up to the vsync's own timestamp and presents at once. A flyback due before a vsync is therefore produced and shown at that vsync, and one due after it cannot be produced until the next, so two flybacks share a refresh only when the browser delivers the callback late. The 10 ms timer tick continues exactly as before, so the sound chip is fed whatever the display does and #885 stays fixed. Fast-forward and fast tape leave the vsync tick to present only. `Display.present` now draws only when a frame has been painted since the last draw, so the loop's call and the display's own animation frame never draw the same frame twice, and the loop passes the frame time so #1135's one-draw-per-frame guard can use it.

## Measured

Same conditions as #1136: 1500 emulated frames per run, two runs.

| | quiet 60 Hz lost | loaded 60 Hz lost | real 30 Hz lost |
|---|---|---|---|
| `main` | 34, 21 | 103, 166 | 600 |
| #1136 alone | 1, 1 | 134, 218 | 600 |
| this PR (on #1136) | 0, 0 | 114 | 601 |

## Trade-offs, plainly

- **Latency under load.** Shown-frame age (flyback to GL upload) in the loaded software-GL run is p50 4.0 ms and p90 12.8 against `main`'s 3.0 and 8.3. I do not have an age measurement for the quiet condition (that harness reads pixels back, which the hardware-GL headless run does not support), so whether this holds on a healthy machine is unmeasured. The mechanism that could cost latency is that a flyback due just after a vsync is now produced by the next timer tick and shown at the following vsync, which is what `main` did too, so I expect the quiet-machine difference to be noise, but that is a prediction.
- **Emulation inside the animation frame.** Up to 16.7 ms of machine can run at the start of a rendering frame, on the main thread, ahead of the present. On a slow machine that pushes the present later within the frame. The timer tick's 100 ms cap still applies.
- **Roughly 60 more advances a second**, each small, on top of the 100 timer ticks. Wakeups are not the concern they were for a 2 ms tick, but it is more work than `main`.
- **No help under load.** When animation-frame callbacks come late or are skipped (rAF p99 of 33 ms on the loaded box) the guarantee lapses and the numbers are `main`'s.
- **With #1135 as well**, the display's `pendingPresent` flag becomes redundant (the queue's emptiness is the state) and should go; the combined build measured 1, 1 lost at quiet 60 Hz and 94, 109 loaded.

## Testing

`tests/unit/test-emulation-loop.js` covers emulating up to the vsync and presenting in the same callback, a vsync a timer tick has already passed, fast-forward, and letting go of the vsyncs once stopped. `tests/unit/test-display.js` covers drawing only when a frame has been painted. Not felt on a real 60 Hz display, and audio underruns under this loop are not measured.


mattgodbolt:
Copilot's suppressed finding on the vsync advance charging callback delay to execute time is right and is fixed in c4a48b2: the vsync timestamp is now only where the machine is emulated to, and the execute, idle and Virtual MHz timings come from when the advance ran. A test pins it. The same commit carries wording and naming changes from a clarity read (`last` is now `emulatedTo`, the display flag is `presented`, the file-top comment no longer claims there is no requestAnimationFrame).

PR #1151, opened 2026-09-16
fix: single-stepping in the debugger no longer leaves a trail of beam dots

mattgodbolt:
Single-stepping left every step's beam dot on the screen (Matt's report, after #1146 landed).

## Cause and fix

The debugger's paint repaints the same frame, so the phosphor's decay count for it was zero and the max blend kept each earlier dot, which is brighter than the black it moved off. A debug paint now replaces the picture, as a restore does: the old picture decays away in full under it. `tests/unit/test-display.js` pins that a debug paint carries the full decay count.

## Checked headless

MODE 2 with the cursor off, the machine started, run a few thousand cycles and stopped into the debugger three times, canvas read back: one dot at the beam, the text intact. On `main` the same sequence on a blank screen leaves a dot per step.

after three steps: one dot


mattgodbolt:
Copilot's one suppressed comment: the rows below the beam are the previous frame's, and every flyback paint refreshes all of them, so a dot from an earlier step can only survive below the beam if that flyback's paint was skipped, which needs a configured `frameSkip` as well as stepping across a frame. Within a frame the beam only moves down, and each debug paint copies the rows down to its dot from a framebuffer the dot has been removed from, so the previous dot is overwritten. Not carrying dot bookkeeping for the frameSkip case; noted here.

PR #1160, opened 2026-09-22
feat: the drive noise follows the head, a click per step and a looped run while it seeks

mattgodbolt:
Draft: the drive noise driven by the head's actual movement, the first tier of the synthesis study. For listening before anything else.

**Today** the noise picks a whole recording by how far a seek command asks the head to go: 1 to 2 tracks the click, 3 to 20 `seek.wav`, 21 to 40 `seek2.wav`, over 40 `seek3.wav`. The recordings are fixed length, so a 30-track seek's 0.72 s of movement gets 1.07 s of sound with 44 clicks, and a 2-track seek's 48 ms gets 148 ms with 4.

**Now** the drive raises a `step` event for every track the head is stepped, and `DdNoise` follows the head:

- one step plays the click;
- a second step within the same movement (60 ms, past the slowest controller rate of 30 ms plus the emulator's 10 ms ticks) stops the click and starts `seek3.wav` looped between click boundaries;
- when the steps stop, the loop is stopped at the next click boundary and the click's ring (from 24 ms into `step.wav`) plays as the settle.

The loop points were measured from the file: the first click begins 5.5 ms in and they come every 24.21 ms, so the loop is 73 intervals from there, joining in the previous click's ring at the same level at both ends. Three constants beside the timeout. The pattern is the one MAME's floppy sound device uses (a step sample, then a looped seek recording chosen by measured interval, then the step sample's tail as the settle), without its pitch shifting.

The per-seek notifications from both controllers go, with the 1770's restore and seek distance reporting from #1158, since nothing else used them.

**Measured** on the Exile v1 boot (drive-noise output recorded headless; movement times from the emulated step rate of 24 ms):

| movement | head moves for | sound before | sound now |
|---|---|---|---|
| 2 tracks | 48 ms | 148 ms, 4 clicks | 98 ms, about 3 clicks |
| 7 tracks | 168 ms | 148 ms | 212 ms |
| 9 tracks | 216 ms | 148 ms | 309 ms |
| 30 tracks | 720 ms | 1066 ms, 44 clicks | 746 ms, about 32 clicks |
| 36-track restore | 864 ms | 1066 ms | 1018 ms |
| single-track steps (title, EXILEL) | | 11 clicks | 11 clicks |

The extra clicks on a run are the 60 ms the loop plays on while waiting to be sure the steps have stopped, about two clicks; the loop cannot be un-stopped once a stop is scheduled, so it has to wait.

**Not validated, by design.** The recording steps at the 8271's 24 ms a track under DFS, and there is no recording at any other rate, so a 1770 stepping at 6 ms still gets clicks 24 ms apart; the run's length is right and its tempo is not. The study found that a fast run is not the single click repeated (the stepper rotor rings at about 180 Hz, so at 6 ms it is driven, not summed), so this wants recordings of a real drive at each rate the emulator uses, or synthesis validated against them. The per-step call here is what either would plug into.

**Listening.** Exile v1 on a BBC B: the catalogue's short seeks, the 30-track seek to the title, four clicks over the title, the restore and seven clicks for EXILEL.


mattgodbolt:
Sounds absoltuely dredaful - the spinning disc sounds seems to start and stop and the sound in general is really bad: the almost-whitenoise in the background of the mechanism rotating (part of the charm) doesn't loop properly (or similar?) Sounds more than that though - like there's a bug in the system, similar to the one we had before you fixed it in the original disc drive system -sounds "restarting" early?

mattgodbolt:
Closing this approach: it decided everything from when step events arrived on the audio clock, and a laptop's emulation hitches made the run end and restart mid-seek. The rework drives the noise from the controller's own seek timing (start with its length and rate, per-step pacing, end at the settle) and schedules the sound on the audio clock from that. New PR to follow.

